Issue #80, opened 2014-01-13
Mock magic methods


armetiz:
Hi there.

Sorry to open this issue, seems that it was already discussed, [here](https://github.com/phpspec/prophecy/issues/13), [here](https://github.com/phpspec/prophecy/issues/34) and [here](https://github.com/phpspec/prophecy/issues/44).

But, right now there is no solution to mock object that didn't define public API using PHP implementations.
And sadly, many libraries uses magic calls, just talk about the example : AWS. I Guess that I can find other popular examples.

[AWS](https://github.com/aws/aws-sdk-php/blob/master/src/Aws/S3/S3Client.php#L99) use PHPDoc to describe API when doing magic functions, see PhpDoc [@method](http://manual.phpdoc.org/HTMLSmartyConverter/PHP/phpDocumentor/tutorial_tags.method.pkg.html).

So, what about using this information to read public API ? PHPStorm use this information to build auto-completion.

Regards,

hackel:
Before I file this as a bug report, is this supposed to work if I add the @method tag to an interface and then mock the interface?  I still get the "x is undefined" error in this case, but if I create a stub class that implements the interface, I can add the @method tag to it, and phpspec picks it up.

jakzal:
@hackel seems very wrong to add an interface with the `@method` tags instead of actual methods. Why would you do that? Interface is a contract, and there's nothing to enforce it if you only add comments in.

stof:
@hackel not reading ``@method`` phpdoc on interfaces was done on purpose (and this has been rejected when asking for it explicitly after that)

hackel:
@jakzal You're right--the only reason I would do it is to get phpspec to be able to mock methods on that interface without complaining with the "undefined" error.

@stof I wasn't able to find another issue discussing this, but why was it rejected?

PHP has no ability to specify that a method in an interface can be implemented in a class via a magic method, so it seems a (non-enforceable) annotation like this is the only way.  I know many people seem to hate magic methods in general, but a lot of packages use them, and I find them incredibly useful.  It allows to have a type of "inheritance" that is actually testable.

I can continue making stub classes and putting my annotations in there, it just seems like an unnecessary step to me.

jakzal:
@hackel think about it. If your interface doesn't have those methods, than it's actually not the right collaborator for your class. You seem to be relying on an implementation. It would break if someone provided an alternative implementation of your interface.

jakzal:
see #169

hackel:
[...]

PR #81, opened 2014-01-13
Add API Discovering patch

armetiz:
Related to this issue #80, I made a PR.

As you can see, I define the @method inside the MagicCallPatch class... It's just for testing purpose and it need to be removed.

But, right now, I don't know how to add functional tests for this feature, or the way I have to do it.. So, please help me on this point.


umpirsky:
To add missing info, in order to cover the issue with magic method call, add `@method missingMethod()` annotation to your class and `method Double\My\Namespace\P20::missingMethod() is not defined` error will go away.

PR #97, opened 2014-03-28
Added support for bypassing a constructor, if it's final

klaussilveira:
When creating a double of a class that has the constructor defined as final, we just ignore it, to avoid a fatal error about overriding a final constructor. An example of this problem occurs when double'ing Phalcon's Model classes.


bendavies:
@everzet @stof is this still the preferred solution? I've come across the problem with Doctrine SQLFilters and will work on a fix if that proposal above is correct.

stof:
@bendavies in any case, you should not be mocking Doctrine's SQLFilters class, unless this is done inside the Doctrine testsuite itself. Otherwise, it means you are testing your own code against what you *think* Doctrine does, not against what it does (the generic rule being "don't mock what you don't own")

bendavies:
I want to assert that a `setParameter` call on my SQLFilter is made, in one of my listeners. How would you achieve this without mocking?

PR #106, opened 2014-05-28
Feature/magic method arguments

vaidasm:
This adds support for arguments in magic methods + solves issue when current or its parents has normal method and also *method* tag (example why its useful would be to get correct return type in PHPStorm)

Argument parsing still lacks typehinting support, but I'm not sure its needed....


everzet:
Can you rebase, please?

PR #118, opened 2014-08-18
Add support for SplFileObject

wodor:
Since the SplFileInfoPatch was opened to allow parameters of subclasses, internal children of SplFileInfo stoped working i.e  code like this 
```
    function it_test_spl_file_object(\SplFileObject $fileObject)
    {
        $this->foo()->shouldReturn('foo');
    }
```

generates 
```
  34  ! it test spl file object
      exception [exc:RuntimeException("SplFileObject::__construct() expects parameter 4 to be resource, null given")] has been thrown.
```

This PR is WIP, missing things: 
- support for `SplTempFileObject` should be added 
- smarter way of letting classes which inherit from `SplFileObject` and `SplTempFileObject`  to have arbitrary parameters in constructor

### DIFF
--- a/spec/Prophecy/Doubler/ClassPatch/SplFileObjectPatchSpec.php
+++ b/spec/Prophecy/Doubler/ClassPatch/SplFileObjectPatchSpec.php
@@ -1,0 +1,71 @@
+<?php
+
+namespace spec\Prophecy\Doubler\ClassPatch;
+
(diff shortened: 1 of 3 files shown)


everzet:
@wodor updates?

sagikazarmark:
Would be nice to have this feature.

PR #127, opened 2014-09-10
Set KeywordPatch to lower priority to check magic methods for keywords

seagoj:
Sets KeywordPatch to a lower priority than MagicCallPatch, so magic methods are filtered for keywords.


everzet:

PR #128, opened 2014-09-12
Add Argument::invoke() for testing callback arguments

tomphp:
The idea here is to cause a callback argument to a mock to be invoked so that the contents of the callback can be tests. Example:

```php
<?php

function it_tests_callback(Collection $collection, Item $item)
{
    $collection->foreachItem(Argument::invoke([$item], 'success'))->willReturn();

    $this->execute($collection, 10);

    $item->action(10)->shouldHaveBeenCalled();
}

function execute($collection, $value)
{
    $collection->foreachItem(function ($item) use ($value) {
        $item->action($value);

        return 'success';
    });
}
```

I understand this could be controversial so I'm expecting some discussion.

Also I wasn't sure what score value to return so I just took the same as `AnyValueToken`?

### DIFF
--- a/src/Prophecy/Argument/Token/InvokeToken.php
+++ b/src/Prophecy/Argument/Token/InvokeToken.php
@@ -1,0 +1,121 @@
+<?php
+
+/*
+ * This file is part of the Prophecy.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *     Marcello Duarte <marcello.duarte@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Prophecy\Argument\Token;
+
+use Prophecy\Util\StringUtil;
+use Prophecy\Exception\InvalidArgumentException;
+
+/**
+ * Invoke callback.
+ *
+ * @author Tom Oram <tom@x2k.co.uk>
+ */
+class InvokeToken implements TokenInterface
+{
+    /**
+     * @var array
+     */
+    private $arguments;
+
+    /**
+     * @var false
+     */
+    private $expectResult;
+
+    /**
+     * @var mixed
+     */
+    private $expectedResult;
+
+    /**
+     * @var StringUtil
+     */
+    private $util;
+
+    /**
+     * @param mixed[] $arguments
+     * @param bool    $expectResult
+     * @param mixed   $expectedResult
+     */
+    public function __construct(array $arguments, $expectResult = false, $expectedResult = null)
+    {
+        $this->arguments      = $arguments;
+        $this->expectResult   = $expectResult;
+        $this->expectedResult = $expectedResult;
+        $this->util           = new StringUtil();
+    }
+
+    /**
+     * Execute the callback with arguments and score 3 (same as AnyValueToken)
+     * if successful.
+     *
+     * @param callable $callback
+     *
+     * @return int|bool
+     */
+    public function scoreArgument($callback)
+    {
+        if (!is_callable($callback)) {
+            throw new InvalidArgumentException(sprintf(
+                'Callable expected as an argument to CallbackToken, but got %s.',
+                gettype($callback)
+            ));
+        }
+
+        $result = call_user_func_array($callback, $this->arguments);
+
+        $score = 3;
+
+        if ($this->expectResult && $result != $this->expectedResult) {
+            $score = false;
+        }
+
+        return $score;
+    }
+
+    /**
+     * Returns false.
+     *
+     * @return bool
+     */
+    public function isLast()
+    {
+        return false;
+    }
+
+    /**
+     * Returns string representation for token.
+     *
+     * @return string
+     */
+    public function __toString()
+    {
+        return sprintf(
+            '%sinvoke(%s)',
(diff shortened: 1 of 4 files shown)


everzet:
@tomphp `thatInvokes` is more inline with other argument checkers

src/Prophecy/Argument/Token/InvokeToken.php
```diff
@@ -0,0 +1,121 @@
...
+     *
+     * @return int|bool
+     */
+    public function scoreArgument($callback)
+    {
+        if (!is_callable($callback)) {
+            throw new InvalidArgumentException(sprintf(
+                'Callable expected as an argument to CallbackToken, but got %s.',
+                gettype($callback)
+            ));
```
stof: shouldn't it return ``false`` when it cannot score an argument instead of throwing an exception ? This exception will break several cases

tomphp: @everzet I have renamed it to `thatInvokes` but not pushed yet so we can continue this conversation. I'm still not 100% sure this is right since what this should really be saying is:

`a callback which, when invoked with these arguments, will return this value`

Thoughts?

@stof I took this from https://github.com/phpspec/prophecy/blob/master/src/Prophecy/Argument/Token/CallbackToken.php#L34 so I'm guessing that if it's incorrect in this PR, then it is there too? Do you mind pointing out where throwing an exception here will cause a problem?

everzet:
After re-reading this PR it seems you're trying to describe very complex, but very unique behaviour. It seems to me that it is so specific that we can question inclusion of such token in the core.

Simply put, I'm really confused by what this thing means:

```php
$collection->foreachItem(Argument::invoke([$item], 'success'))->willReturn();
```

And I think this thing doesn't make it clearer:

```php
$collection->foreachItem(Argument::callbackReceivingAndReturning([$item], 'success'))->willReturn();
```

That's probably because I don't have a full context about your problem. And that's another hint that this token is too specific for your spec domain.

tomphp:
## TL;DR
I've been thinking about this a bit more.

Ultimately, there are other way's to achieve the behaviour I'm discussing here. I do feel that there are times where the approach I'm talking about is a neater/better way of doing things. However, as a *highly opinionated* framework, it think it's fair enough for prophecy to encourage one of the other ways (even if I disagree ;) ).

If you feel this is outside the scope of the core then feel free to close the PR. (in which case I may make it a self contained library).

## More Thoughts
The *usual* way of achieving this behaviour is using a `foreach` like so:

```php
public function execute(Collection $collection)
{
    foreach ($collection->getItems() as $item) {
        $item->performAction();
    }
}
```

If similar actions across the content of the collection are performed in many places in the codebase, then you end up with a lot of tests which might look like this:

```php
function it_perform_action_across_all_items(Collection $coll, Item $item1, Item $item2)
{
    $coll->getItems()->willReturn([$item1, $item2]);

    $this->execute($coll);

    $item1->performAction()->shouldHaveBeenCalled();
    $item2->performAction()->shouldHaveBeenCalled();
}
```

This gets pretty tiresome, and if the body of the loop is more complex then the tests get worse.

If the use of the collection has a limited scope, then obviously a better approach is have a domain specific method on the collection:

```php
class Collection
{
    private $items = [];

    function performActionAcrossItems()
    {
        foreach ($this->items as $item) {
            $item->performAction();
        }
    }
}
```

```php
public function execute(Collection $collection)
{
    $collection->performActionAcrossItems();
}
```

This way the tests are simple:

```php
function it_perform_action_across_all_items(Collection $coll)
{
    $this->execute($coll);

    $coll->performActionAcrossItems()->shouldHaveBeenCalled();
}
```

This approach follows the idea of *tell don't ask*.

However, if the scope of actions to be performed on the items is unknown, this last method is not possible. This is when a methods like `foreachItem` become useful (also things like map, reduce, filter also make a good fit):


```php
class Collection
{
    private $items = [];

    function foreachItem(Closure $callback)
    {
        foreach ($this->items as $item) {
            $callback($item);
        }
    }
}
```

```php
public function execute(Collection $collection)
{
    $collection->foreachItem(function ($item) {
        $item->performAction();
    });
}
```

But currently this is untestable (may be possible though a convoluted used of `Argument::that()` but doesn't even bare thinking about as it would be indecipherable). Therefore, AFAIK the best we can do is:

```php
function it_perform_action_across_all_items(Collection $coll)
{
    $this->execute($coll);

    $coll->foreachItem(Argument::type('Closure'))->shouldHaveBeenCalled();
}
```

I firmly believe there is a place for such code - and even consider it *good code*. However, as I stated earlier, maybe it's just not in the style of prophecy?

jon-acker:
@tomphp my argument against this formulation would be that Argument:: is primarily a *matcher*, as such people expect to use it as a matcher and don't expect it to execute methods on a collaborator.

You're looking for a shortcut to validate that your $items have been operated on. I don't think that writing the spec the way you suggest makes this clear.

what prevents you from doing:
```php
    foreach ($collection as $item) {
        $item->performAction()->shouldHaveBeenCalled();
    }
```

everzet:
@tomphp your example is where I'd consider not using mocks for that edge case. Mocks as any other tool have limitation. I think you're trying to stretch them too thin.

tomphp:
Fair enough and thanks for considering.

Issue #139, opened 2015-01-14
Fatal error when mocking reference returning interface methods 


edsonmedina:
Example:

    interface A
    {
        public function &foo ($x, $y, $z);
    }
    
    $prophet = new Prophecy\Prophet;
    
    $x = $prophet->prophesize();
    $x->willImplement ('A');
    $z = $x->reveal();

This will throw a PHP fatal error:

    PHP Fatal error:  Declaration of Double\A\P1::blah() must be compatible with & A::blah($x, $y, $z) in /home/edson/tmp/vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(49) : eval()'d code on line 2
    PHP Stack trace:
    PHP   1. {main}() /home/edson/tmp/test.php:0
    PHP   2. Prophecy\Prophecy\ObjectProphecy->reveal() /home/edson/tmp/test.php:14
    PHP   3. Prophecy\Doubler\LazyDouble->getInstance() /home/edson/tmp/vendor/phpspec/prophecy/src/Prophecy/Prophecy/ObjectProphecy.php:105
    PHP   4. Prophecy\Doubler\Doubler->double() /home/edson/tmp/vendor/phpspec/prophecy/src/Prophecy/Doubler/LazyDouble.php:122
    PHP   5. Prophecy\Doubler\Doubler->createDoubleClass() /home/edson/tmp/vendor/phpspec/prophecy/src/Prophecy/Doubler/Doubler.php:105
    PHP   6. Prophecy\Doubler\Generator\ClassCreator->create() /home/edson/tmp/vendor/phpspec/prophecy/src/Prophecy/Doubler/Doubler.php:142
    PHP   7. eval() /home/edson/tmp/vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php:49
    
    shell returned 255

edsonmedina:
Possibly related to this?
https://bugs.php.net/bug.php?id=48019

aik099:
@edsonmedina , I recommend you adding `Closes #1234` in the PR description instead of mentioning that issue in the commit message itself. This way GitHub will properly link PR to the issue.

PR #140, opened 2015-01-15
Added support for methods that return reference (Issue #139)

edsonmedina:


aik099:
Closes #139

src/Prophecy/Doubler/Generator/Node/MethodNode.php
```diff
@@ -71,6 +72,16 @@ public function setStatic($static = true)
         $this->static = (bool) $static;
     }
 
+    public function returnsReference()
+    {
+        return $this->returnsReference;
+    }
+
+    public function setReturnsReference($value = true)
```
stof: the argument should not be called ``$value``

stof:
I see several issues here:

- this should be covered by specs (and btw, the current specs are failing because of the extra calls you added)
- I'm quite sure that this does not fix the issue entirely, because the method code will not be compatible with returning by reference AFAICT
- in any case, mocking a function returning by ref will not give you a reference to anything else than a local variable inside the double. doing something else would require Prophecy to return by reference everywhere (which would create weird issues for all other cases)

edsonmedina:
* Specs fixed.
* How does it currently deal with arguments passed by reference?

stof:
@edsonmedina they are not reference anymore inside the Prophecy layer, so you cannot change it by reference

edsonmedina:
I agree that this does not solve the issue completely, but at least it doesn't break anymore when mocking a class/interface with a reference returning method. Currently it dies even if you're not touching that specific method.

My patch won't allow you to return references, but it will allow you to extend the class/interface and mock other stuff. Which is better than nothing IMO.

stof:
But it will still break if you call the method

stof:
the ClassMirrorSpec and the ClassGeneratorSpec are still missing specs about the behavior of functions returning by references (if you revert your ClassGenerator change, the spec will still pass)

edsonmedina:
The reason why i'm fixing this is I'm trying to mock an external dependency on a project. The dependency has a method (which is not called at all) that returns by reference. Just by extending the class I was getting a fatal error (even if I'm never going to use that method). 

Sure, if I call the method now, weird things will happen.

So what do you suggest? Maybe throwing an exception when calling the method?

(I'll add specs for the new methods' behaviour)

everzet:
@edsonmedina throwing an exception sounds fine by me.

edsonmedina:
I think I screwed up with a rebase, I'll do another clean push request, sorry.

everzet:
Okay :)

PR #141, opened 2015-01-16
Use correct case when only 1 prediction was made

dantleech:
Currently:

````
    but 1 were made:
      - flush() @ src/Sulu/Component/HttpCache/EventSubscriber/KernelSubscriber.php:109
````

with this PR:

````
    but 1 was made:
      - flush() @ src/Sulu/Component/HttpCache/EventSubscriber/KernelSubscriber.php:109
````


sstok:
:+1:

everzet:

PR #151, opened 2015-01-18
Fixed invalid version string defined for doctrine/instantiator

dspasic:


GrahamCampbell:
:-1: This is not a fix. The version is valid. You need to update composer on your system.

dspasic:
I apologies and thank you for the hint!

PR #152, opened 2015-01-19
add support for magic calls argument parsing

vaidasm:
Rebased #106

### DIFF
--- a/spec/Prophecy/Doubler/ClassPatch/MagicCallPatchSpec.php
+++ b/spec/Prophecy/Doubler/ClassPatch/MagicCallPatchSpec.php
@@ -58,6 +58,7 @@ class MagicCallPatchSpec extends ObjectBehavior
         $method->addArgument(new ArgumentNode('param1'));
         $method->addArgument(new ArgumentNode('second'));
 
+        $node->addMethod(Argument::any())->willReturn(null);
         $node->addMethod($method)->shouldBeCalled();
 
         $this->apply($node);
@@ -72,15 +73,16 @@ class MagicCallPatchSpec extends ObjectBehavior
 
         $method = new MethodNode('withHintedType');
 
-        $argument = new ArgumentNode('integer');
-        $argument->setDefault("2");
-        $argument->setTypeHint("int");
+        $argument = new ArgumentNode('floatVar');
+        $argument->setDefault("2.3");
+        $argument->setTypeHint("float");
         $method->addArgument($argument);
 
         $argument = new ArgumentNode('class');
         $argument->setTypeHint('\spec\Prophecy\Doubler\ClassPatch\MagicalApi');
         $method->addArgument($argument);
 
+        $node->addMethod(Argument::any())->willReturn(null);
         $node->addMethod($method)->shouldBeCalled();
 
         $this->apply($node);
@@ -107,6 +109,7 @@ class MagicCallPatchSpec extends ObjectBehavior
         $argument->setDefault('\DateTime::RSS');
         $method->addArgument($argument);
 
+        $node->addMethod(Argument::any())->willReturn(null);
         $node->addMethod($method)->shouldBeCalled();
 
         $this->apply($node);
@@ -135,7 +138,7 @@ class MagicalApi
 /**
  * @method void simpleParam($param1,  $second )
  * @method void withDefaultValue($singleQuote = 'value', $doubleQuote="value", $const= \DateTime::RSS)
- * @method void withHintedType(int $integer = 2, \spec\Prophecy\Doubler\ClassPatch\MagicalApi $class)
+ * @method void withHintedType(float $floatVar = 2.3, \spec\Prophecy\Doubler\ClassPatch\MagicalApi $class)
  */
 class MagicalParametrizedApi
 {
--- a/spec/Prophecy/Doubler/ClassPatch/MagicCallPatchSpec.php
+++ b/spec/Prophecy/Doubler/ClassPatch/MagicCallPatchSpec.php
@@ -54,12 +54,59 @@ class MagicCallPatchSpec extends ObjectBehavior
     {
         $node->getParentClass()->willReturn('spec\Prophecy\Doubler\ClassPatch\MagicalParametrizedApi');
 
-        $method = new MethodNode('parametrizedMethod');
-        $method->addArgument(new ArgumentNode('param'));
-        $argumentWithDefaultValue = new ArgumentNode('param');
-        $argumentWithDefaultValue->setDefault('value');
-        $method->addArgument($argumentWithDefaultValue);
-        $method->addArgument($argumentWithDefaultValue);
+        $method = new MethodNode('simpleParam');
+        $method->addArgument(new ArgumentNode('param1'));
+        $method->addArgument(new ArgumentNode('second'));
+
+        $node->addMethod($method)->shouldBeCalled();
+
+        $this->apply($node);
+    }
+
+    /**
+     * @param \Prophecy\Doubler\Generator\Node\ClassNode $node
+     */
+    function it_discovers_api_with_hinted_parameters_using_phpdoc($node)
+    {
+        $node->getParentClass()->willReturn('spec\Prophecy\Doubler\ClassPatch\MagicalParametrizedApi');
+
+        $method = new MethodNode('withHintedType');
+
+        $argument = new ArgumentNode('integer');
+        $argument->setDefault("2");
+        $argument->setTypeHint("int");
+        $method->addArgument($argument);
+
+        $argument = new ArgumentNode('class');
+        $argument->setTypeHint('\spec\Prophecy\Doubler\ClassPatch\MagicalApi');
+        $method->addArgument($argument);
+
+        $node->addMethod($method)->shouldBeCalled();
+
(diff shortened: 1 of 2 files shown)


.idea/.name
```diff
@@ -0,0 +1 @@
+prophecy
```
sstok: This must be reverted, please add `.idea/*` in your `.git/info/exclude`

vaidasm:
Sorry for .idea. Removed that

spec/Prophecy/Doubler/ClassPatch/MagicCallPatchSpec.php
```diff
@@ -54,6 +73,7 @@ function it_has_50_priority()
 
 /**
  * @method void undefinedMethod()
+ * @method void definedMethod()
```
stof: why changing this class ? It looks wrong to me

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
```diff
@@ -55,12 +56,36 @@ public function apply(ClassNode $node)
...
+
+            $node->addMethod($methodNode);
         }
     }
 
     /**
+     * @param array $argument
+     * @return ArgumentNode
+     */
+    protected function parseArgument(array $argument) {
```
stof: this should be a private method (and the curly brace should be on its own line)

spec/Prophecy/Doubler/ClassPatch/MagicCallPatchSpec.php
```diff
@@ -46,6 +47,24 @@ function it_ignores_existing_methods($node)
...
+    function it_discovers_api_with_parameters_using_phpdoc($node)
+    {
+        $node->getParentClass()->willReturn('spec\Prophecy\Doubler\ClassPatch\MagicalParametrizedApi');
+
+        $method = new MethodNode('parametrizedMethod');
+        $method->addArgument(new ArgumentNode('param'));
+        $argumentWithDefaultValue = new ArgumentNode('param');
+        $argumentWithDefaultValue->setDefault('value');
+        $method->addArgument($argumentWithDefaultValue);
+        $method->addArgument($argumentWithDefaultValue);
```
everzet: Why calling it two times?

spec/Prophecy/Doubler/ClassPatch/MagicCallPatchSpec.php
```diff
@@ -46,6 +47,24 @@ function it_ignores_existing_methods($node)
...
+        $node->addMethod($method)->shouldBeCalled();
+
+        $this->apply($node);
```
everzet: What will happen if you remove all the logic exception `$nod->addMethod($method)` from the `MagicCallPatch` implementation? I bet example will still pass and I bet that's not what you expect.

everzet:
Sorry for earlier comments. Mornings...

everzet:
Could you refactor `MagicCallPatch::apply()`? You complicated it a bit:
https://scrutinizer-ci.com/g/phpspec/prophecy/inspections/8ea3e5a5-a873-43c2-875a-dad2da87b5bc

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
```diff
@@ -55,12 +56,37 @@ public function apply(ClassNode $node)
...
+
                 $node->addMethod($methodNode);
             }
+
+            $node->addMethod($methodNode);
         }
     }
 
     /**
+     * @param array $argument
```
stof: can you document the type of data in the array (probably ``string[]`` here, right ?)

spec/Prophecy/Doubler/ClassPatch/MagicCallPatchSpec.php
```diff
@@ -67,10 +86,19 @@ public function definedMethod()
...
  * @method void undefinedMethod()
  * @method void definedMethod()
  */
 class MagicalApiExtended extends MagicalApi
 {
 
-}
\ No newline at end of file
+}
+
```
stof: the extra line at the end should be removed

spec/Prophecy/Doubler/ClassPatch/MagicCallPatchSpec.php
```diff
@@ -67,10 +86,19 @@ public function definedMethod()
 }
 
 /**
+ * @method void parametrizedMethod($param, $param = 'value', $param = "value")
```
stof: using ``$param`` several times is wrong. Arguments must have different names

spec/Prophecy/Doubler/ClassPatch/MagicCallPatchSpec.php
```diff
@@ -46,6 +47,24 @@ function it_ignores_existing_methods($node)
...
+    {
+        $node->getParentClass()->willReturn('spec\Prophecy\Doubler\ClassPatch\MagicalParametrizedApi');
+
+        $method = new MethodNode('parametrizedMethod');
+        $method->addArgument(new ArgumentNode('param'));
+        $argumentWithDefaultValue = new ArgumentNode('param');
+        $argumentWithDefaultValue->setDefault('value');
+        $method->addArgument($argumentWithDefaultValue);
+        $method->addArgument($argumentWithDefaultValue);
+        $node->addMethod($method)->shouldBeCalled();
```
stof: Can you separate this call from the building of ``$method`` to make it more readable ?

spec/Prophecy/Doubler/ClassPatch/MagicCallPatchSpec.php
stof: there is 2 missing things to support:

- typed arguments (e.g. ``\stdClass $param``)
- arguments defining their default value without space around the ``=`` (which may not be parsed properly by the phpDocumentor library from my quick look).

vaidasm:
Ok so I did a minor refactoring on argument parsing. Also added typehinted argument support.

One question is what should we do if parsing method tag failed? throw exception? ignore? sth else?

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
```diff
@@ -49,14 +52,56 @@ public function apply(ClassNode $node)
...
+        return $methodNode;
+    }
+
+    /**
+     * @param string $argument
+     * @return ArgumentNode
+     */
+    private function parseArgument($argument)
+    {
+        if (preg_match('/^([\w\\\\]+)?\s*\$?(\w+)(?:\s*=\s*[\'\"]?([\w\\\\:]+)[\'\"]?)?/', $argument, $matches)) {
```
stof: ``\w+`` does not represent the allowed chars in a PHP variable name

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
```diff
@@ -49,14 +52,56 @@ public function apply(ClassNode $node)
...
+            $argumentNode = new ArgumentNode($matches[2]);
+
+            if (!empty($matches[1])) {
+                $argumentNode->setTypeHint($matches[1]);
             }
+
+            if (!empty($matches[3])) {
+                $argumentNode->setDefault($matches[3]);
```
stof: your handling of the default looks wrong. It can be lots of other things than strings (integers, booleans, null, float)

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
```diff
@@ -49,14 +52,56 @@ public function apply(ClassNode $node)
...
+            }
+
+            return $argumentNode;
+        } else {
```
stof: use an early throw instead of wrapping everything in the if/else

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
```diff
@@ -69,5 +114,4 @@ public function getPriority()
     {
         return 50;
     }
-}
-
+}
```
stof: I asked to remove the extra empty line at the end, but you should keep the final EOL

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
```diff
@@ -49,14 +52,56 @@ public function apply(ClassNode $node)
...
+                $method = $this->parseMethod($tag);
 
-            if (!$reflectionClass->hasMethod($methodName)) {
-                $methodNode = new MethodNode($tag->getMethodName());
-                $methodNode->setStatic($tag->isStatic());
+                if (!$reflectionClass->hasMethod($method->getName())) {
+                    $node->addMethod($method);
+                }
+            } catch(RuntimeException $e) {
+                // What happens we cannot parse method tag? should we issue a warning here?
```
stof: you should catch this around the handling of arguments IMO, so that the method is still added without its arguments (same than the current behavior)

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
stof: I asked to remove the extra empty line at the end, but you should keep the final EOL

PR #155, opened 2015-01-25
Record arguments used when method not found exception is thrown

ciaranmcnulty:
This will be useful when handling these exceptions in PhpSpec (e.g. by generating a method)

### DIFF
--- a/spec/Prophecy/Exception/Doubler/MethodNotFoundExceptionSpec.php
+++ b/spec/Prophecy/Exception/Doubler/MethodNotFoundExceptionSpec.php
@@ -9,7 +9,7 @@ class MethodNotFoundExceptionSpec extends ObjectBehavior
 {
     function let()
     {
-        $this->beConstructedWith('', 'User', 'getName');
+        $this->beConstructedWith('', 'User', 'getName', array(1,2,3));
     }
 
     function it_is_DoubleException()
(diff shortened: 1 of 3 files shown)


spec/Prophecy/Exception/Doubler/MethodNotFoundExceptionSpec.php
```diff
@@ -9,7 +9,7 @@ class MethodNotFoundExceptionSpec extends ObjectBehavior
 {
     function let()
     {
-        $this->beConstructedWith('', 'User', 'getName');
+        $this->beConstructedWith('', 'User', 'getName', array(1,2,3));
```
GrahamCampbell: Is the cs intentionally incorrect here? `array(1, 2, 3)` is what we want isn't it?

src/Prophecy/Exception/Doubler/MethodNotFoundException.php
```diff
@@ -15,18 +15,24 @@ class MethodNotFoundException extends DoubleException
 {
     private $classname;
     private $methodName;
+    /**
+     * @var array|null
```
stof: this should always be an array, not null IMO

src/Prophecy/Exception/Doubler/MethodNotFoundException.php
```diff
@@ -15,18 +15,24 @@ class MethodNotFoundException extends DoubleException
...
+    private $arguments;
 
     /**
      * @param string $message
      * @param string $classname
      * @param string $methodName
+     * @param array  $arguments
      */
-    public function __construct($message, $classname, $methodName)
+    public function __construct($message, $classname, $methodName, $arguments = array())
```
stof: please typehint the new argument as array

src/Prophecy/Prophecy/MethodProphecy.php
```diff
@@ -46,9 +46,10 @@ public function __construct(ObjectProphecy $objectProphecy, $methodName, $argume
     {
         $double = $objectProphecy->reveal();
         if (!method_exists($double, $methodName)) {
+            var_dump(2,$arguments);
```
stof: this must be removed

ciaranmcnulty: eeeesh, what an amateur

everzet:
Oh LOL

PR #160, opened 2015-02-11
ReturnArgumentPromise should accept optional index argument

coderbyheart:
This adds an optional zero-based integer argument to `->willReturnArgument()` which allows
to specify the argument which should be returned.

The default behaviour of `ReturnArgumentPromise` is not changed, it will return the first
argument by default.

### DIFF
--- a/src/Prophecy/Promise/ReturnArgumentPromise.php
+++ b/src/Prophecy/Promise/ReturnArgumentPromise.php
@@ -25,24 +25,24 @@ class ReturnArgumentPromise implements PromiseInterface
     /**
      * @var int
      */
-    private $argument;
+    private $index;
 
     /**
      * Initializes callback promise.
      *
-     * @param int $argument The zero-indexed number of the argument to return
+     * @param int $index The zero-indexed number of the argument to return
      *
      * @throws \Prophecy\Exception\InvalidArgumentException
      */
-    public function __construct($argument = 0)
+    public function __construct($index = 0)
     {
-        if (!is_int($argument) || $argument < 0) {
+        if (!is_int($index) || $index < 0) {
             throw new InvalidArgumentException(
                 'Zero-based index expected as argument to ReturnArgumentPromise, but got %s.',
-                $argument
+                $index
             );
         }
-        $this->argument = $argument;
+        $this->index = $index;
     }
 
     /**
@@ -56,6 +56,6 @@ class ReturnArgumentPromise implements PromiseInterface
      */
     public function execute(array $args, ObjectProphecy $object, MethodProphecy $method)
     {
-        return count($args) > $this->argument ? $args[$this->argument] : null;
+        return count($args) > $this->index ? $args[$this->index] : null;
     }
 }
--- a/src/Prophecy/Promise/ReturnArgumentPromise.php
+++ b/src/Prophecy/Promise/ReturnArgumentPromise.php
@@ -11,6 +11,7 @@
 
 namespace Prophecy\Promise;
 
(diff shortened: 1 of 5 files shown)


src/Prophecy/Promise/ReturnArgumentPromise.php
```diff
@@ -22,7 +23,30 @@
 class ReturnArgumentPromise implements PromiseInterface
 {
     /**
-     * Returns first argument if has one, null otherwise.
+     * @var int
+     */
+    private $argN;
```
stof: Please don't use an abbreviation (``$argumentIndex`` or just ``$argument`` might be better names)

coderbyheart: Thanks for the feedback. I've renamed it in 202a79f79f66f8ee8443e3565fee03d35115579.

sstok:
:+1:

src/Prophecy/Promise/ReturnArgumentPromise.php
```diff
@@ -22,7 +23,30 @@
...
+     */
+    public function __construct($argument = 0)
+    {
+        if (!is_int($argument) || $argument < 0) {
+            throw new InvalidArgumentException(
+                'Zero-based index expected as argument to ReturnArgumentPromise, but got %s.',
+                $argument
+            );
+        }
+        $this->argument = $argument;
```
everzet: I'd call it `index`

coderbyheart: @everzet Should I update the PR?

everzet: yes

coderbyheart: @everzet Done on f665cd2f3365da5b7f83dc3df911d898ccaff507

coderbyheart:
Thanks a lot!

ciaranmcnulty:
Very nice

src/Prophecy/Promise/ReturnArgumentPromise.php
```diff
@@ -32,6 +56,6 @@ class ReturnArgumentPromise implements PromiseInterface
      */
     public function execute(array $args, ObjectProphecy $object, MethodProphecy $method)
     {
-        return count($args) ? $args[0] : null;
+        return count($args) > $this->index ? $args[$this->index] : null;
```
docteurklein: no existence check ?

coderbyheart: $args is guaranteed to be a non-associative array and index is an int. So the existence check is implicit in the condition.

docteurklein: yes but if index (let's say `2` does not exist, it will trigger a notice. Shouldn't we transform this into a proper exception ?

coderbyheart: A user could set `$this->index` to `-1` then an `E_NOTICE` would occur  which would be converted to a failed test anyway.

stof: @coderbyheart no you cannot. Setting it to -1 would throw an exception in the constructor.

@docteurklein the count condition already prevents the notice

coderbyheart: @stof Right, I forgot that …

docteurklein: oh right! Sorry for the noise.

PR #166, opened 2015-03-11
Avoiding abandoned packages "dflydev/markdown"

xphere:
As `phpdocumentor/reflection-docblock` depends on the abandoned package `dflydev/markdown` until 2.0.1, this causes a warning in `composer update --prefer-lowest`.

The problem is my projects depends on `phpunit/phpunit: ~4.5` which depends on `phpspec/prophecy: ~1.3.1` which depends on `phpdocumentor/reflection-docblock: ~2.0` which depends on the abandoned package `dflydev/markdown`… Would this be enough to fix that chain?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         }
     ],
     "require": {
-        "phpdocumentor/reflection-docblock": "~2.0",
+        "phpdocumentor/reflection-docblock": "~2.0, >=2.0.1",
         "sebastian/comparator":              "~1.1",
         "doctrine/instantiator":             "^1.0.2"
     },


stof:
Well, I don't think this is needed actually. We don't have any reason to actually forbid using phpdocumentor/reflection-docblock 2.0.1 in Prophecy (it works fine with our code too).
and btw, an abandonned package does not mean it does not work anymore. It means it is not maintained anymore. So for a ``--prefer-lowest`` install, it is not an issue IMO (you should of course never use the ``--prefer-lowest`` option elsewhere than in a CI build)

xphere:
Well, I'm developing directly with `--prefer-lowest` and that makes me feel more comfortable with dependencies, but I can live with that warning. What annoys me more is precisely to have a red banner in my travis builds. Maybe that's something I should get used to.

I believe it's better to fix this in the root, if possible, than to have another direct dependency in every project depending on prophecy for testing. Even though it's a simple `composer require --dev "phpdocumentor/reflection-docblock: ~2.0, >=2.0.1"`

stof:
If we add a dev requirement in Prophecy, it won't help at all for your own Travis builds (dev requirements are not used outside the root package)

composer.json
```diff
@@ -17,7 +17,7 @@
         }
     ],
     "require": {
-        "phpdocumentor/reflection-docblock": "~2.0",
+        "phpdocumentor/reflection-docblock": "~2.0, >=2.0.1",
```
GrahamCampbell: would need to be `^2.0.1`

xphere:
@stof The `dev` dependency is only for our projects: I don't need prophecy nor phpdocumentor in my prod environment. Prophecy should require phpdocumentor with no `dev`, as it does, but update the lowest supported version.

@GrahamCampbell I didn't know this notation at the time, thanks for pointing out

everzet:
@xphere if your project depends on `>=2.0.1`, then it is your project `composer.json` responsibility to state that, not Prophecy.

PR #167, opened 2015-03-11
Fix stringifying of hash containing one value

avant1:
Equivalence operator is not strict enough.
The following code
```php
var_dump([0 => 0] == [0 => 'foo']);
```
outputs "true".

So, when we are trying to stringify ```['foo' => 'bar']```, hash is being stingified as array.


everzet:

PR #169, opened 2015-03-13
Discover even Interface magic API not only Class

jakubriedl:
MagicCallPatch ambition is to discover magic APIs. It does it well on classes but ignores interfaces and that is a problem when you mock them.

### DIFF
--- a/src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
+++ b/src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
@@ -41,9 +41,23 @@ class MagicCallPatch implements ClassPatchInterface
      */
     public function apply(ClassNode $node)
     {
-        $parentClass = $node->getParentClass();
-        $reflectionClass = new \ReflectionClass($parentClass);
+        $this->attach($node, $node->getParentClass());
 
+        $interfaces = $node->getInterfaces() ?: [];
+        foreach ($interfaces as $interfaceName) {
+            $this->attach($node, $interfaceName);
+        }
+    }
+
+    /**
+     * Discovers and attaches Magical API to node
+     *
+     * @param ClassNode $node
+     * @param $className
+     */
+    protected function attach(ClassNode $node, $className)
+    {
+        $reflectionClass = new \ReflectionClass($className);
         $phpdoc = new DocBlock($reflectionClass->getDocComment());
 
         $tagList = $phpdoc->getTagsByName('method');
@@ -51,7 +65,7 @@ class MagicCallPatch implements ClassPatchInterface
         foreach($tagList as $tag) {
             $methodName = $tag->getMethodName();
 
-            if (!$reflectionClass->hasMethod($methodName)) {
+            if (!$node->hasMethod($methodName)) {
                 $methodNode = new MethodNode($tag->getMethodName());
                 $methodNode->setStatic($tag->isStatic());
 
--- a/src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
+++ b/src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
@@ -43,7 +43,7 @@ class MagicCallPatch implements ClassPatchInterface
     {
         $this->attach($node, $node->getParentClass());
 
-        $interfaces = $node->getInterfaces() ?: [];
+        $interfaces = $node->getInterfaces() ?: array();
         foreach ($interfaces as $interfaceName) {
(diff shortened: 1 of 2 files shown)


stof:
Magic API does not make sense on an interface, given that an interface cannot enforce implementations to implement these magic methods

jakubriedl:
Role of Interfaces are to define API of their implementators and enforce it when possible.
Consider this scenario:

- I have system based on separated components
- All components have shared Contracts with defined Interfaces
- I have Interface for query lang. It has 50 methods with similar action (store the call for later usage)
- I don't want to force implementator to have 50 methods in their class but I need to tell that the methods should be callable
- So it should be done using phpdoc which is integrated with everything
- And at the final I need to mock any implementator in other components in test. And there I don't know which one will be used

stof:
@jakubriedl but you will have no guarantee that the implementor of the interface actually define these magic methods. They are not part of the contract enforced by the interface.

@everzet what do you think about this ?

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
```diff
@@ -41,9 +41,23 @@ public function supports(ClassNode $node)
...
+        }
+    }
+
+    /**
+     * Discovers and attaches Magical API to node
+     *
+     * @param ClassNode $node
+     * @param $className
+     */
+    protected function attach(ClassNode $node, $className)
```
stof: this should not be protected. There is no reason to make it an extension point.

And ``attach`` does not describe what the method does. ``registerMagicMethods`` would be a better name

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
```diff
@@ -41,9 +41,23 @@ public function supports(ClassNode $node)
...
+        foreach ($interfaces as $interfaceName) {
+            $this->attach($node, $interfaceName);
+        }
+    }
+
+    /**
+     * Discovers and attaches Magical API to node
+     *
+     * @param ClassNode $node
+     * @param $className
```
stof: your phpdoc is invalid given that it misses the type before the parameter name

jakubriedl:
No I don't have the guarantee but as well I don't have guarantee that it is implemented well. So I use specBDD tests to verify that.
In this scenario I use Interfaces just for definition.

everzet:
Parseable doc-blocks are not part of your public API. Interface, virtual or not, with 50 methods screams for [ISP](http://en.wikipedia.org/wiki/Interface_segregation_principle) violation. You have a design pain of "this interface has too many methods" and instead of tackling this problem with composition you want Prophecy to help you hide the symptoms.

Problem is that Prophecy's goal is to expose design pains, not hide them.

jakubriedl:
I understand that Prophecy's goal is to expose design pains, thats the reason why I use it.

ISP is not violated in this scenario because all methods have same purpose. Methods in this case are for example `where`, `orWhere`, `whereNot`, `whereNotNull`, `orWhereNotBetween` ... Currently I have 25 of them but in future there this number can grow

everzet:
So why not having them explicitly on an interface then?

jakubriedl:
Because every method have same functionality -> save the call for later use. It is typical case when to use __call method. Also I cannot force implementator to write 200 (8*25) more lines which are completely useless because that is code smell.

This pull request doesn't force anyone to use this but enables the usage.

I agree that in most cases developer should avoid this. But in some like this or when you have lot of separated components you don't have any other choice.

everzet:
> Also I cannot force implementator to write 200 (8*25) more lines which are completely useless because that is code smell. It is typical case when to use __call method.

You don't want implementor to write 50 useless methods, but you want him to constantly keep all of them in his head? You have a **classical** example of sponge-interface - interface that absorbs all the logic around it. Monstrous interface with 50 methods should be refactored into hierarchy of much simpler objects, not hidden behind dynamic dispatch. In your particular case I'd look into QueryBuilder pattern instead of `__call` fiddling.

> This pull request doesn't force anyone to use this but enables the usage.

Prophecy is a very opinionated framework. Having choice of doing things the dirty way is not the choice I'm interested in supporting. Unless you prove that this is **the only way**, which is not the case ATM.

jakubriedl:
My **example is not sponge-interface** because its basically one method with many forms. orWhere has absolutely same functionality as where. Name is just expression of form.

And **this approach is the only way** how to solve my issue. I need to create mock based on interface because I don't have any implementation in project. It is injected from another in DI.

Issue #181, opened 2015-04-18
[Bug] As of 1.4, multiple callback expectations no longer work


WouterJ:
```php
$errorHandler = $prophecy->prophesize('Acme\ErrorHandler');

$errorHandler->handle(Argument::that(function ($error) {
    return $error->getSeverity() === 'error';
})->shouldBeCalled();

$errorHandler->handle(Argument::that(function ($error) {
    return $error->getSeverity() === 'warning';
})->shouldBeCalled();

$errorHandler->handle(new Error(..., Error::ERROR));
$errorHandler->handle(new Error(..., Error::WARNING));
```

The above code passes when using 1.3.1, but fails in 1.4.0 with:
```
method call:
  - trigger(Error Object (
    ...
    'severity' => 'warning'
  ))
  on ... was not expected, expected calls were:
    - trigger(callback())
```

When debugging, you can find that only the first expectation is checked each time.

everzet:
@WouterJ sorry for late answer. Can you verify that #183 indeed fixes your problem?

WouterJ:
Issue was fixed :)

PR #183, opened 2015-04-25
Fix for #181

everzet:


WouterJ:
Thanks @everzet, this fixed the problem

PR #188, opened 2015-05-20
Feature:  boundary prediction.

ranpafin:
This will implement a simple boundary prediction such as *at least* and *no more than*  via:
- shouldHaveBeenCalledAtLeastTimes
- shouldHaveBeenCalledNoMoreThanTimes

Specs included to ensure correct behavior.

Regards,
Francesco.

### DIFF
--- a/spec/Prophecy/Prediction/LowerBoundaryCallPredictionSpec.php
+++ b/spec/Prophecy/Prediction/LowerBoundaryCallPredictionSpec.php
@@ -1,0 +1,70 @@
+<?php
+
+namespace spec\Prophecy\Prediction;
+
+use PhpSpec\ObjectBehavior;
+use Prophecy\Prophecy\MethodProphecy;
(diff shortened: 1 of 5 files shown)


sstok:
:+1: for this feature.

taueres:
Good feature!
It would be nice to support non-spying predictions too.

Adding the following methods should do the trick:
 * `MethodProphecy#shouldBeCalledAtLeastTimes($times)`
 * `MethodProphecy#shouldBeCalledNoMoreThanTimes($times)`

everzet:
I'm generally against checking amount of times method was called. I'm still very torn on having `shouldBeCalledTimes()` in the core. Checking boundary conditions in my opinion is taking it one step too far.

Having your own interaction that you deeply care about (otherwise why doubling?), but not knowing how many exact calls are there feels weird to me.

You can always achieve this behaviour with custom predictions, so I'm closing this PR.

PR #190, opened 2015-05-27
Return self promise

Taluu:
Sometimes, some chains are made on an object we would like to mock. Currently, there is no way AFAIK to be able to return the revealed object when calling one of its method.

With this patch, it should be able to do so. I am not too sure about what I did for the spec of `ReturnSelfPromise`, as I never did any phpspec before...

### DIFF
--- a/src/Prophecy/Promise/ReturnSelfPromise.php
+++ b/src/Prophecy/Promise/ReturnSelfPromise.php
@@ -1,0 +1,38 @@
+<?php
+
+/*
+ * This file is part of the Prophecy.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *     Marcello Duarte <marcello.duarte@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Prophecy\Promise;
+
+use Prophecy\Prophecy\ObjectProphecy;
+use Prophecy\Prophecy\MethodProphecy;
+
+/**
+ * Return self promise.
+ *
+ * @author Baptiste Clavié <clavie.b@gmail.com>
+ */
+class ReturnSelfPromise implements PromiseInterface
+{
+    /**
+     * Returns the object prophecy revealed
+     *
+     * @param array          $args
+     * @param ObjectProphecy $object
+     * @param MethodProphecy $method
+     *
+     * @return ObjectProphecy
+     */
+    public function execute(array $args, ObjectProphecy $object, MethodProphecy $method)
+    {
+        return $object;
+    }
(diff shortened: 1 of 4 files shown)


Taluu:
Hello,

any news on that ? Thanks.

weierophinney:
It *is* possible to do currently, I just discovered:

```php
$fluent = $this->prophesize('Something\Fluent');
$fluent->withSomething('interesting')->will(function () use ($fluent) {
    return $fluent->reveal();
});
```

It's a bit verbose, but it works.

stof:
@weierophinney well, everything possible with custom promise implementations is also possible with a callback, as the callback does the full Promise implementation.

Taluu:
But now I'm not quite sure if I should reveal the object in the `execute` method (rather than just the prophecy ?), as it will be revealed later anyway IRC

Basster:
+1 for this.

But I'm also not shure about revealing the object in the `execute`. In my tests, both ways (with and without `->reveal()`) working properly.

src/Prophecy/Promise/ReturnSelfPromise.php
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * @param array          $args
+     * @param ObjectProphecy $object
+     * @param MethodProphecy $method
+     *
+     * @return object
+     */
+    public function execute(array $args, ObjectProphecy $object, MethodProphecy $method)
+    {
+        return $object->reveal();
```
stof: revealing is useless. The value returned by a promise is already revealed by Prophecy

Taluu: Ok this confirm my thought https://github.com/phpspec/prophecy/pull/190#issuecomment-130398366, i'll amend that.

Taluu: fixed

everzet:
I'd rather not encourage mocking of fluent interfaces. I'm happy with having solution outside of the core.

aik099:
In fact this also worked for me:
```php
$fluent = $this->prophesize('Something\Fluent');
$fluent->withSomething('interesting')->willReturn($fluent);
```

stof:
I think that the current possibility is fine enough

teohhanhui:
@aik099

Why does it work? (I tried looking at the code but can't quite figure it out yet...)

Intuitively I'd expect:

```php
$fluent = $this->prophesize('Something\Fluent');
$fluent->withSomething('interesting')->willReturn($fluent->reveal());
```

aik099:
@teohhanhui it works, because object is automatically revealed where needed. Maybe your code would also work correctly.

PR #191, opened 2015-05-27
Add a Lazy Reveal return promise

Taluu:
This allows a mock call on a method to return a mocked object, and to reveal it at the latest time possible, instead of having to reveal it at the method declaration.

```php
<?php
// ... uses, class init, ... etc

$mock = $this->prophesize();
$mock2 = $this->prophesize();

$mock2->getMock()->will(new Promise\LazyRevealReturnPromise($mock));
// instead of
$mock2->getMock()->willReturn($mock->reveal());
```

I was not sure if I should add or not a helper in the `MethodProphecy`, as this is kind of specific (and how to name it : `willLazilyReturns()` ? `willLazilyRevealsAndReturn()` ?

### DIFF
--- a/spec/Prophecy/Promise/LazyRevealReturnPromiseSpec.php
+++ b/spec/Prophecy/Promise/LazyRevealReturnPromiseSpec.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace spec\Prophecy\Promise;
+
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+use Prophecy\Prophecy\ObjectProphecy;
+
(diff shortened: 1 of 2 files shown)


stof:
``$mock2->getMock()->willReturn($mock)`` already does it. Prophecy will reveal any value returned by the promise before returning it: https://github.com/phpspec/prophecy/blob/5a355f91730c845301a9e28f91c8a5053353c496/src/Prophecy/Prophecy/ObjectProphecy.php#L186-191

Taluu:
oh, good point, I didn't see that. All is well then. :)

PR #200, opened 2015-07-11
Add support for scalar parameter types and return types

trowski:
This PR adds support for PHP 7 scalar parameter types and return types while maintaining backward compatibility with older PHP versions.

When a method prophecy is created for a method with a return type, a `CallbackPromise` is automatically added to the method prophecy that will return an appropriate type. The user can override this promise, but it prevents a method call from throwing `TypeError` if a return promise is not specified.

I'm using this library as part of PHPUnit. With this patch I was able to successfully test PHP 7 code that was otherwise failing due to: 1) revealed mocks not implementing an interface because of missing parameter or return types; or 2) methods returning `null` instead of the declared return type, resulting in a `TypeError` being thrown.

### DIFF
--- a/src/Prophecy/Doubler/Generator/Node/MethodNode.php
+++ b/src/Prophecy/Doubler/Generator/Node/MethodNode.php
@@ -126,6 +126,19 @@ class MethodNode
                 $this->returnType = $type;
                 break;
 
+            case 'double':
+            case 'real':
+                $this->returnType = 'float';
+                break;
+
+            case 'boolean':
+                $this->returnType = 'bool';
+                break;
+
+            case 'integer':
+                $this->returnType = 'int';
+                break;
+
             default:
                 $this->returnType = '\\' . ltrim($type, '\\');
         }
--- a/src/Prophecy/Doubler/Generator/Node/MethodNode.php
+++ b/src/Prophecy/Doubler/Generator/Node/MethodNode.php
@@ -25,6 +25,7 @@ class MethodNode
     private $visibility = 'public';
     private $static = false;
     private $returnsReference = false;
+    private $returnType;
 
     /**
      * @var ArgumentNode[]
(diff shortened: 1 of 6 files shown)


src/Prophecy/Doubler/Generator/Node/MethodNode.php
```diff
@@ -100,6 +101,41 @@ public function getArguments()
...
+    {
+        switch ($type) {
+            case null:
+            case '':
+                $this->returnType = null;
+                break;
+
+            case 'string';
+            case 'float':
+            case 'int':
```
GrahamCampbell: What about the aliases like integer, double, real?

trowski: I based this switch on what can result from `(string) ReflectionMethod::getReturnType()`. If this function can be called by the user, then I can certainly add aliases that map to the correct type.

trowski:
@everzet @stof Thoughts on merging this PR? I've submitted similar patches for PHPUnit and Mockery, which have been merged. I'm just waiting on Prophecy so I can do some proper testing of PHP 7 code.

everzet:
@trowski almost all your tests are failing.

everzet:
Closed in favour of complete version in #204

trowski:
@everzet That's odd that all the tests would be failing. Everything worked fine when I created the PR, did something change in-between that caused most of the tests to fail?

src/Prophecy/Doubler/Generator/Node/MethodNode.php
ciaranmcnulty: FYI `int`, `float`, `string` and `bool` are valid scalar type hints, `integer`, `double` and `real` are not

trowski: Right, I questioned adding the aliases but I figured it wouldn't hurt anything. Feel free to delete them, since as far as I can tell this method will only be called using the result from `ReflectionMethod::getReturnType()`, which of course will only return the valid scalar types.

PR #203, opened 2015-08-04
[Bugfix] Prevent final methods from being manually extended.

kamioftea:
If the DisableConstructorPatch is registered with the doubler, and a class with a final __construct is doubled, then 
 * The __construct method is not added to the ClassNode as it is final
 * The DisableConstructorPatch detects that the ClassNode has no __construct, so adds one
 * When eval() is called on the generated code, the extended final class causes a FatalError.

This pull request:
 * Marks final classes as un-extendable.
 * Throws an exception if a method that is un-extendable is added to the class node
 * ~~The DisableConstructorPatch checks if __construct is extendable before adding it~~

**EDIT:** DisableConstructorPatch changes are no longer part of this PR as it leads to undesirable behavior.


src/Prophecy/Doubler/Generator/Node/ClassNode.php
```diff
@@ -23,6 +23,7 @@ class ClassNode
     private $parentClass = 'stdClass';
     private $interfaces  = array();
     private $properties  = array();
+    protected $unextendableMethods = [];
```
stof: should be private.

and please don't use the short array syntax. This library is compatible with PHP 5.3 (as seen by the Travis failure)

src/Prophecy/Doubler/Generator/Node/ClassNode.php
```diff
@@ -100,6 +101,12 @@ public function getMethods()
 
     public function addMethod(MethodNode $method)
     {
+        if(!$this->isExtendable($method->getName()))
+        {
```
everzet: Curly bracket should be together with the previous line and there should be space between `if` and `(`

src/Prophecy/Doubler/Generator/Node/ClassNode.php
```diff
@@ -127,4 +134,33 @@ public function hasMethod($name)
...
+        return $this->unextendableMethods;
+    }
+
+    /**
+     * @param string $unextendableMethod
+     */
+    public function addUnextendableMethod($unextendableMethod)
+    {
+        if(!$this->isExtendable($unextendableMethod))
+        {
```
everzet: Same as above

src/Prophecy/Doubler/Generator/Node/ClassNode.php
```diff
@@ -100,6 +101,12 @@ public function getMethods()
 
     public function addMethod(MethodNode $method)
     {
+        if(!$this->isExtendable($method->getName()))
+        {
+            throw new \InvalidArgumentException(sprintf(
```
everzet: You should throw custom exception here instead of generic.

src/Prophecy/Doubler/Generator/Node/ClassNode.php
```diff
@@ -127,4 +134,33 @@ public function hasMethod($name)
...
+    public function addUnextendableMethod($unextendableMethod)
+    {
+        if(!$this->isExtendable($unextendableMethod))
+        {
+            return;
+        }
+        $this->unextendableMethods[] = $unextendableMethod;
+    }
+
+	/**
```
GrahamCampbell: indentation

everzet:
With this solution Prophecy will still "double" classes with final constructor, would it? It would simple ignore constructor and create a double of the class with original constructor. If so, that's very unpredictable and hence incorrect behaviour.

kamioftea:
Yes it would use the code from the original constructor.

I've reverted the changes to the DisableConstructorPatch So this PR now just adds the logic for adding a method that can't be extended, as this gives a better error than the test suite exiting with a Fatal Error.

everzet:
Love it. Thank you very much for contribution!

PR #204, opened 2015-08-13
Add support for scalar parameter types and return types

everzet:
Finalisation of #200

### DIFF
--- a/spec/Prophecy/Doubler/Generator/ClassMirrorSpec.php
+++ b/spec/Prophecy/Doubler/Generator/ClassMirrorSpec.php
@@ -53,9 +53,11 @@ class ClassMirrorSpec extends ObjectBehavior
         $method2->getParameters()->willReturn(array());
         $method3->getParameters()->willReturn(array());
 
-        $method1->hasReturnType()->willReturn(false);
-        $method2->hasReturnType()->willReturn(false);
-        $method3->hasReturnType()->willReturn(false);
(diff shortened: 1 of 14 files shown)


everzet:
@trowski thank you very much for contribution!

spec/Prophecy/Prophecy/MethodProphecySpec.php
```diff
@@ -4,6 +4,11 @@
 
 use PhpSpec\ObjectBehavior;
 
+class ClassWithFinalMethod
```
stof: are you sure about this change ? I thought PhpSpec would not run files properly if the spec is not the first class in it

spec/Prophecy/Doubler/Generator/ClassMirrorSpec.php
```diff
@@ -86,6 +92,10 @@ function it_changes_argument_names_if_they_are_varying($class, $method, $paramet
         $method->isStatic()->willReturn(false);
         $method->returnsReference()->willReturn(false);
 
+        if (version_compare(PHP_VERSION, '7.0', '>=')) {
```
lisachenko: I think, that checking of PHP_VERSION_ID can be faster and more transparent.

trowski:
@everzet You're welcome. Sorry that there was some things to clean up, that wasn't my intention. `MethodNode::hasReturnType()` should have always returned `false` in any tests that would be run under 5.x and that there would be no need for all the version comparisons. Not sure where that went wrong...

PR #218, opened 2015-10-15
Added missing sprintf for an exception message in ReturnArgumentPromise.php

mgrajcarek:
If $index is for example an Object, what happend to me once during a test, fatal error is raised. Second parameter is a $code which should be an integer.
Also, from the Exception message construction (%s), it can be assumed that that's what an author wanted to do.


stof:
Well, for an integer, the value should still be displayed in the message, otherwise it is not helpful

everzet:
Was fixed with 6273d145

PR #219, opened 2015-10-16
Added @var for IDE-typehinting

edsonmedina:
PHPStorm complains about variable not being a callable

### DIFF
--- a/src/Prophecy/Prophecy/MethodProphecy.php
+++ b/src/Prophecy/Prophecy/MethodProphecy.php
@@ -86,7 +86,7 @@ class MethodProphecy
                     case 'Traversable':
                     case 'Generator':
                         // Remove eval() when minimum version >=5.5
-                        /** @var callable  */
+                        /** @var callable $generator */
                         $generator = eval('return function () { yield; };');
                         return $generator();
 
--- a/src/Prophecy/Prophecy/MethodProphecy.php
+++ b/src/Prophecy/Prophecy/MethodProphecy.php
@@ -86,6 +86,7 @@ class MethodProphecy
                     case 'Traversable':
                     case 'Generator':
                         // Remove eval() when minimum version >=5.5
+                        /** @var callable  */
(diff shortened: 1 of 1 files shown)


src/Prophecy/Prophecy/MethodProphecy.php
```diff
@@ -86,6 +86,7 @@ public function __construct(ObjectProphecy $objectProphecy, $methodName, $argume
                     case 'Traversable':
                     case 'Generator':
                         // Remove eval() when minimum version >=5.5
+                        /** @var callable  */
```
stof: missing variable name

edsonmedina: I'm just being consistent. 
The other places in the code using @var type-hints don't use variable names either.

edsonmedina: I can add it if you prefer

src/Prophecy/Prophecy/MethodProphecy.php
```diff
@@ -86,6 +86,7 @@ public function __construct(ObjectProphecy $objectProphecy, $methodName, $argume
                     case 'Traversable':
                     case 'Generator':
                         // Remove eval() when minimum version >=5.5
+                        /** @var callable $generator */
                         $generator = eval('return function () { yield; };');
```
dawehner: Is it just me, or is the minimum version 5.5 already, when you use generators?

aik099: If you look up, then this code is wrapped with an IF statement where PHP 7.0 or above is required.

ciaranmcnulty:
Why add a @var when it's immediately returned anyhow?

jakzal:
@ciaranmcnulty what's returned is a result of calling the generator, not the generator itself. 

The way PHPStorm  complains looks like this:

phpspec-phpstorm

Not sure if this is a big deal. Endusers won't see it.

Issue #121, opened 2014-09-06
Replace Prophecy\Util\ExportUtil with sebastianbergmann/exporter


whatthejeff:
The [`Prophecy\Util\ExportUtil`](https://github.com/phpspec/prophecy/blob/0f713ca01bfc5d19065596a16435fb75bf007eba/src/Prophecy/Util/ExportUtil.php) class should be replaced by [`sebastianbergmann/exporter`](https://github.com/sebastianbergmann/exporter/blob/master/src/Exporter.php). Not sure if there's a clean way to do this without breaking backwards compatibility since prophecy's version has some small customizations, but `ExportUtil` is missing bug fixes, performance improvements, copyright attributions, tests, etc. which are present in @sebastianbergmann's version.

PR #226, opened 2015-11-02
Updated ExportUtil to reflect the latest version of sebastianbergmann…

### LINKED ISSUE #221: ExportUtil::recursiveExport Allowed memory size x bytes exhausted
I'm mocking an call on a prophecy that uses a doctrine object as parameter. like this: 

$prophecy = $this->prophesize(ClassToMock::class);
$prophecy->aMethodCall($doctrineEntity)->shouldBeCalledTimes(1);

Everything is fine until we make the checkPredictions check, and when the test fails (I havent made the required amount of calls to the prophecy) the expected output is printed to screen using ExportUtil, right? 
Now, a doctrine entity might be highly recursive, and when outputting it using print_r as exportUtil does, this results in a never ending loop and the php stalls to memory exhaust. 

I think there should be some more sophisticated way to print the output of the object. For example something on the style of doctrine/common Debug::dump?

jakari:
…/exporter

Fixes #221 and #121 
Relates to phpspec/phpspec#785

### DIFF
--- a/src/Prophecy/Util/ExportUtil.php
+++ b/src/Prophecy/Util/ExportUtil.php
@@ -3,7 +3,7 @@
 namespace Prophecy\Util;
 
 use Prophecy\Prophecy\ProphecyInterface;
-use SplObjectStorage;
+use SebastianBergmann\RecursionContext\Context;
 
 /*
  * This file is part of the Prophecy.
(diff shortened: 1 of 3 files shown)


jakari:
Ping @stof @everzet any thoughts? One possibility of course could be to use the library..?

src/Prophecy/Util/ExportUtil.php
```diff
@@ -113,73 +124,88 @@ protected static function recursiveExport($value, $indentation, &$processedObjec
...
-            if (strlen($content) > 0) {
-                $content = "\n" . $content . $whitespace;
+            return sprintf('Array &%s (%s)', $key, $values);
+        }
+
+        if (is_object($value)) {
+            $class = get_class($value);
+
+            if ($value instanceof ProphecyInterface) {
+                return sprintf('%s Object (*Prophecy*)', get_class($value));
```
sstok: You can reuse the `$class` variable.

everzet:

PR #231, opened 2015-11-20
Fix #91: Variadic Resubmission

### LINKED ISSUE #91: Mocking variadic functions
PHP 5.6 introduces variadic arguments for functions. Prophecy should be updated to support mocking such methods.

Note that I might provide a patch (unless someone beats be). I'm opening the issue here as a reminder

SERVANT14:
Following up from #143. I'm a newbie to phpSpec, just would really like variadic's supported. Let me know what needs to be done to get this in.

There are three failures in the test suite, all with this message:
```
- it generates proper php code for specific ClassNode
- it generates proper php code for variadics
- it overrides properly methods with args passed by reference

      method call:
        - isOptional()
      on Double\Prophecy\Doubler\Generator\Node\ArgumentNode\P221 was not expected, expected calls were:
        - getName()
        - getTypeHint()
        - hasDefault()
        - getDefault()
        - isPassedByReference()
        - isVariadic()
```

Should this be "isVariadic()" rather than "isOptional()"?

### DIFF
--- a/spec/Prophecy/Doubler/Generator/ClassCodeGeneratorSpec.php
+++ b/spec/Prophecy/Doubler/Generator/ClassCodeGeneratorSpec.php
@@ -54,20 +54,23 @@ class ClassCodeGeneratorSpec extends ObjectBehavior
 
         $argument11->getName()->willReturn('fullname');
         $argument11->getTypeHint()->willReturn('array');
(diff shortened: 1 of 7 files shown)


everzet:
Travis fails.

PR #229, opened 2015-11-18
Don't call debug_backtrace when isn't strictly required

felixcarmona:
the makeCall is asking for the debug_backtrace() even when it's a dummy and dummies doesn't have file/line.

it also causes segmentation faults some times, when prophecy tries to call the debug_backtrace on a false object which is being destructed (and the garbage collector hasn't destructed it yet)

```
   20.4270   33866304                                                 -> Double\XXXX\P363->__destruct() /XXXX/vendor/phpspec/phpspec/src/PhpSpec/Wrapper/Subject/Caller.php:287
   20.4271   33866336                                                   -> Double\XXXX\P363->getProphecy() /XXXX/vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(49) : eval()'d code:29
   20.4271   33866432                                                   -> func_get_args() /XXXX/vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(49) : eval()'d code:29
   20.4271   33866520                                                   -> Prophecy\Prophecy\ObjectProphecy->makeProphecyMethodCall() /XXXX/vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(49) : eval()'d code:29
   20.4271   33866520                                                     -> PhpSpec\Wrapper\Unwrapper->reveal() /XXXX/vendor/phpspec/prophecy/src/Prophecy/Prophecy/ObjectProphecy.php:188
   20.4271   33866520                                                       -> PhpSpec\Wrapper\Unwrapper->unwrapOne() /XXXX/vendor/phpspec/phpspec/src/PhpSpec/Wrapper/Unwrapper.php:68
   20.4271   33866568                                                         -> is_array() /XXXX/vendor/phpspec/phpspec/src/PhpSpec/Wrapper/Unwrapper.php:42
   20.4271   33867048                                                         -> array_map() /XXXX/vendor/phpspec/phpspec/src/PhpSpec/Wrapper/Unwrapper.php:43
   20.4271   33866656                                                     -> Prophecy\Call\CallCenter->makeCall() /XXXX/vendor/phpspec/prophecy/src/Prophecy/Prophecy/ObjectProphecy.php:189
   20.4272   33866704                                                       -> debug_backtrace() /XXXX/vendor/phpspec/prophecy/src/Prophecy/Call/CallCenter.php:57
```

```
Program received signal SIGSEGV, Segmentation fault.
[...]

PR #232, opened 2015-11-23
Exclude 'args' in the generated backtrace

ockcyp:
This removes 'args' element from generated backtraces. If arguments are objects it will result in great performance benefit.

See [PHP: debug_backtrace - Manual](http://php.net/manual/en/function.debug-backtrace.php)

People with PHP version less than 5.3.6 will not be able to get any benefit from this change.

This is also a possible fix to #229
It fixed the random segmentation faults we were having in our PHPUnit tests.

gdb backtrace of the segmentation fault:
```
#0  debug_backtrace_get_args (curpos=<value optimized out>)
    at /usr/src/debug/php-5.5.30/Zend/zend_builtin_functions.c:2029
```

### DIFF
--- a/src/Prophecy/Call/CallCenter.php
+++ b/src/Prophecy/Call/CallCenter.php
@@ -54,7 +54,17 @@ class CallCenter
      */
     public function makeCall(ObjectProphecy $prophecy, $methodName, array $arguments)
     {
-        $backtrace = debug_backtrace();
+        // For efficiency exclude 'args' from the generated backtrace
+        if (PHP_VERSION_ID >= 50400) {
+            // Limit backtrace to last 3 calls as we don't use the rest
+            // Limit argument was introduced in PHP 5.4.0
+            $backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 3);
+        } elseif (defined('DEBUG_BACKTRACE_IGNORE_ARGS')) {
+            // DEBUG_BACKTRACE_IGNORE_ARGS was introduced in PHP 5.3.6
+            $backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
+        } else {
+            $backtrace = debug_backtrace();
+        }
 
         $file = $line = null;
         if (isset($backtrace[2]) && isset($backtrace[2]['file'])) {


src/Prophecy/Call/CallCenter.php
```diff
@@ -54,7 +54,13 @@ public function __construct(StringUtil $util = null)
      */
     public function makeCall(ObjectProphecy $prophecy, $methodName, array $arguments)
     {
-        $backtrace = debug_backtrace();
+        // For efficiency leave out 'args' in the generated backtrace
+        // DEBUG_BACKTRACE_IGNORE_ARGS was introduced in PHP 5.3.6
+        if (version_compare(PHP_VERSION, '5.3.6', '>=')) {
```
aik099: You can check for constant presence instead.

src/Prophecy/Call/CallCenter.php
```diff
@@ -54,7 +54,13 @@ public function __construct(StringUtil $util = null)
...
+            $backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
```
aik099: Maybe you'd better exclude `DEBUG_BACKTRACE_PROVIDE_OBJECT` option as well.

ockcyp: @aik099 I excluded that one when I provided `DEBUG_BACKTRACE_IGNORE_ARGS`, overriding default behaviour.

see:
```php
<?php

class A
{
    public function a()
    {
        return debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
    }

    public function b()
    {
        return debug_backtrace();
    }

    public function c()
    {
        return debug_backtrace(DEBUG_BACKTRACE_PROVIDE_OBJECT);
    }
}

$a = new A;
print_r($a->a());
print_r($a->b());
print_r($a->c());
```

output:
```
Array
(
    [0] => Array
        (
            [file] => /home/omer/Projects/Amara/prophecy/backtraceDebug.php
            [line] => 22
            [function] => a
            [class] => A
            [type] => ->
        )
)
Array
(
    [0] => Array
        (
            [file] => /home/omer/Projects/Amara/prophecy/backtraceDebug.php
            [line] => 23
            [function] => b
            [class] => A
            [object] => A Object
                (
                )
            [type] => ->
            [args] => Array
                (
                )
        )
)
Array
(
    [0] => Array
        (
            [file] => /home/omer/Projects/Amara/prophecy/backtraceDebug.php
            [line] => 24
            [function] => c
            [class] => A
            [object] => A Object
                (
                )
            [type] => ->
            [args] => Array
                (
                )
        )
)
```

aik099: Great.

src/Prophecy/Call/CallCenter.php
ockcyp: @aik099 you're right. done.

stof:
What about limiting the length of the returned array too ? We don't need the full backtrace

ockcyp:
great idea! I'll add that in.

ockcyp:
@stof Done.

src/Prophecy/Call/CallCenter.php
```diff
@@ -54,7 +54,17 @@ public function __construct(StringUtil $util = null)
      */
     public function makeCall(ObjectProphecy $prophecy, $methodName, array $arguments)
     {
-        $backtrace = debug_backtrace();
+        // For efficiency exclude 'args' from the generated backtrace
+        if (version_compare(PHP_VERSION, '5.4.0', '>=')) {
```
stof: use ``PHP_VERSION_ID`` to check the PHP version instead

ockcyp:
@stof done.

ockcyp:
Thanks. When can we expect a release including this change?

PR #235, opened 2015-12-03
Fix code generation for scalar parameters

trowski:
The latest builds of PHP 7 forbid a `\` preceding scalar types, causing a fatal error. This PR fixes the code generation of parameters with scalar type declarations.

### DIFF
--- a/src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
+++ b/src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
@@ -73,10 +73,18 @@ class ClassCodeGenerator
             $php = '';
 
             if ($hint = $argument->getTypeHint()) {
-                if ('array' === $hint || 'callable' === $hint) {
-                    $php .= $hint;
-                } else {
-                    $php .= '\\'.$hint;
+                switch ($hint) {
+                    case 'array':
+                    case 'callable':
+                    case 'string':
+                    case 'int':
+                    case 'float':
+                    case 'bool':
+                        $php .= $hint;
+                        break;
+
+                    default:
+                        $php .= '\\'.$hint;
                 }
             }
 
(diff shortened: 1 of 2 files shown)


klaussilveira:
Just experienced this. :+1:

xurumelous:
+1

nesl247:
Please get this merged in.

mathop:
+1

aik099:
This change would be a BC break for projects, that use classes named as scalar types.

avant1:
Probably check for PHP version could be added to avoid BC break (scalar type names are reserved keywords in php7 and cannot be used as class names).

aik099:
Yes, PHP version check is a good solution. There is some code already that checks for PHP version to create PHP version-specific mock code.

ferodss:
+1 for checking PHP version to avoid BC break

williancarminato:
:+1:

yokomizor:
:+1:

trowski:
Good call, I didn't consider classes with scalar type names in the root namespace. It's not a good idea, but I've seen it done…

I added a couple version checks so that the old behavior is maintained in PHP 5.x, but the proper scalar type names are produced in PHP 7.

danilofreitas:
:+1:

everzet:

src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
```diff
@@ -60,7 +60,9 @@ private function generateMethod(Node\MethodNode $method)
             $method->returnsReference() ? '&':'',
             $method->getName(),
             implode(', ', $this->generateArguments($method->getArguments())),
-            $method->hasReturnType() ? sprintf(': %s', $method->getReturnType()) : ''
+            version_compare(PHP_VERSION, '7.0', '>=') && $method->hasReturnType()
```
stof: this should not be necessary IMO. We should never have a return type in the MethodNode on older versions

PR #236, opened 2015-12-09
Approximate value

dantleech:
This value adds an ApproximateValueToken.

The approximate value rounds both the given and argument value to the given precision and compares them, returning 10 in case of a match.

This would seem to be useful in the case of expecting floating point numbers.

### DIFF
--- a/src/Prophecy/Argument.php
+++ b/src/Prophecy/Argument.php
@@ -195,4 +195,18 @@ class Argument
     {
         return new Token\IdenticalValueToken($value);
     }
+
+    /**
+     * Check that argument is same value when rounding to the
+     * given precision.
+     *
+     * @param float $value
+     * @param float $precision
+     *
+     * @return Token\ApproximateValueToken
+     */
+    public static function approximate($value, $precision = 0)
+    {
+        return new Token\ApproximateValueToken($value, $precision);
+    }
 }
--- a/src/Prophecy/Argument/Token/ApproximateValueToken.php
+++ b/src/Prophecy/Argument/Token/ApproximateValueToken.php
@@ -1,0 +1,55 @@
+<?php
+
+/*
+ * This file is part of the Prophecy.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *     Marcello Duarte <marcello.duarte@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Prophecy\Argument\Token;
+
+/**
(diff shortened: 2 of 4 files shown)


src/Prophecy/Argument.php
```diff
@@ -195,4 +195,17 @@ public static function is($value)
     {
         return new Token\IdenticalValueToken($value);
     }
+
+    /**
+     * Check that argument is approximately the same value.
+     *
+     * @param float $value
+     *
```
dantleech: Add precision doc.

src/Prophecy/Argument.php
```diff
@@ -195,4 +195,17 @@ public static function is($value)
...
+     * @return Token\ApproximateValueToken
+     */
+    public static function approximate($value, $precision)
+    {
+        return new Token\ApproximateValueToken($value, $precision);
+    }
+
```
everzet: Extra space

dantleech:
updated

src/Prophecy/Argument.php
```diff
@@ -195,4 +195,18 @@ public static function is($value)
...
+    /**
+     * Check that argument is same value when rounding to the
+     * given precision.
+     *
+     * @param float $value
+     * @param float $precision
+     *
+     * @return Token\ApproximateValueToken
+     */
+    public static function approximate($value, $precision)
```
stof: the precision should be optional here, to match the token constructor (people are likely to use this helper more often than instantiating the token directly btw)

src/Prophecy/Argument/Token/ApproximateValueToken.php
```diff
@@ -0,0 +1,55 @@
...
+        $this->value = $value;
+        $this->precision = $precision;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function scoreArgument($argument)
+    {
+        return round($argument, $this->precision) == round($this->value, $this->precision) ? 10 : false;
```
stof: should use a strict comparison

src/Prophecy/Argument.php
dantleech: I wonder what the default should be then, I suppose it makes most sense to round it to a whole number, i.e. percision = 0 by default.

stof: deciding on the default needs to be discussed (I agree that using ``0`` may be sensible), but I'm sure it should be the same than in the token itself.

dantleech:
Set the precision default to 0 on both factory method and constructor, assuming that 0 is the best default.

everzet:

PR #239, opened 2015-12-10
Fixed missing sprintf in InvalidArgumentException __construct call

emmanuelballery:
Having this fatal error ``PHP Fatal error:  Wrong parameters for Exception([string $exception [, long $code [, Exception $previous = NULL]]])``, I guess this exception is missing a ``sprintf``.


jakzal:
:+1:

PR #241, opened 2015-12-19
Enhance exception for invalid argument to ThrowPromise

Tobion:
Currently the exception is not really useful with something like `Exception class or instance expected as argument to ThrowPromise, but got string`


everzet:

PR #242, opened 2015-12-19
Fix phpdoc for magic methods

Tobion:


everzet:

PR #244, opened 2015-12-23
Specify the required php version for composer

### LINKED ISSUE #224: Specify minimal needed version in "composer.json"
Right now `composer.json` says that minimal required PHP version is PHP 5.3, but in `description` key only. It would be better to write this in `require` key as well.

jakzal:
Fixes #224

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,7 @@
         }
     ],
     "require": {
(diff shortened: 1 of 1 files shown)


sstok:
:+1:

PR #245, opened 2015-12-28
Document consecutive return feature

mbrodala:
See https://gist.github.com/gquemener/292e7c5a4bbb72fd48a8#gistcomment-1222753

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ This promise will cause any call to our double's `read()` method with exactly on
 argument - `'123'` to always return `'value'`. But that's only for this
 promise, there's plenty others you can use:
 
-- `ReturnPromise` or `->willReturn(1)` - returns a value from a method call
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
Minot sure we should encourage this

everzet:
This functionality is intentionally not exposed.

mbrodala:
@everzet And what is the intention if I may ask? :-)

aik099:
Maybe because it allows to do the `structure binding` (preffered way in Mockery) instead of doing `message binding` (preffered way in Prophecy) as described in http://everzet.com/post/72910908762/conceptual-difference-between-mockery-and-prophecy.

If I'm correct, then using `->willReturn(1, 3, 4, ...)` technique would make your test suite aware of tested code internals thus making it more brittle.

mbrodala:
@aik099 Makes perfect sense, thanks. I'll watch out for this.

PR #247, opened 2016-01-13
PhpDoc for interfaces should be used too

ImmRanneft:
look at this code:

https://github.com/nrk/predis/blob/master/src/ClientInterface.php

When you want to test code with Predis/Client you can't do it because of method is not defined error.
This pull request closes it,


ImmRanneft:
No one intersted in this patch?

aik099:
Looks like maintainers of the repo are busy with something else at the moment (nothing gets merged recently).

According to http://www.phpdoc.org/docs/latest/references/phpdoc/tags/method.html this is absolutely valid use case to use `@method` on the interface and it does seem to be inline with Prophecy philosophy as well.

everzet:

PR #253, opened 2016-02-08
add support for 'self' and 'parent' return type

bendavies:
When a php7 return type is `self`, like so:

```
class Foo {
  public function setBar(Bar $bar) : self
  {
      ...
      return $this;
  }
}
```
Extending `Foo::setBar` like so
```
class Baz extends Foo {
  public function setBar(Bar $bar) : self
  {
      ...
  }
}
```
will result in
```
Compile Error: Declaration of Baz::setBar(Bar $bar): Baz must be compatible with Foo::setBar(Bar $bar): Foo
```
So the extending class must respect the return type:
```
class Baz extends Foo {
  public function setBar(Bar $bar) : Foo
  {
      ...
  }
}
```

### DIFF
--- a/spec/Prophecy/Doubler/Generator/ClassMirrorSpec.php
+++ b/spec/Prophecy/Doubler/Generator/ClassMirrorSpec.php
@@ -2,9 +2,8 @@
 
 namespace spec\Prophecy\Doubler\Generator;
 
-use PhpSpec\ObjectBehavior;
 use I\Simply;
-
+use PhpSpec\ObjectBehavior;
 use ReflectionClass;
 use ReflectionMethod;
 use ReflectionParameter;
@@ -13,15 +12,17 @@ class ClassMirrorSpec extends ObjectBehavior
 {
     /**
      * @param ReflectionClass  $class
+     * @param ReflectionClass  $parent
      * @param ReflectionMethod $method1
      * @param ReflectionMethod $method2
      * @param ReflectionMethod $method3
      */
     function it_reflects_a_class_by_mirroring_all_its_public_methods(
-        $class, $method1, $method2, $method3
+        $class, $parent, $method1, $method2, $method3
     )
     {
         $class->getName()->willReturn('Custom\ClassName');
+        $class->getParentClass()->willReturn($parent);
         $class->isInterface()->willReturn(false);
         $class->isFinal()->willReturn(false);
         $class->getMethods(ReflectionMethod::IS_ABSTRACT)->willReturn(array());
@@ -29,9 +30,15 @@ class ClassMirrorSpec extends ObjectBehavior
             $method1, $method2, $method3
         ));
 
+        $parent->getName()->willReturn('Custom\ParentClassName');
+
+        $method1->getDeclaringClass()->willReturn($class);
+        $method2->getDeclaringClass()->willReturn($class);
+        $method3->getDeclaringClass()->willReturn($class);
+
         $method1->getName()->willReturn('getName');
-        $method2->getName()->willReturn('isPublic');
-        $method3->getName()->willReturn('isAbstract');
+        $method2->getName()->willReturn('getSelf');
+        $method3->getName()->willReturn('getParent');
 
         $method1->isFinal()->willReturn(false);
         $method2->isFinal()->willReturn(false);
@@ -54,9 +61,12 @@ class ClassMirrorSpec extends ObjectBehavior
         $method3->getParameters()->willReturn(array());
 
         if (version_compare(PHP_VERSION, '7.0', '>=')) {
-            $method1->hasReturnType()->willReturn(false);
-            $method2->hasReturnType()->willReturn(false);
-            $method3->hasReturnType()->willReturn(false);
+            $method1->hasReturnType()->willReturn(true);
+            $method1->getReturnType()->willReturn('string');
+            $method2->hasReturnType()->willReturn(true);
+            $method2->getReturnType()->willReturn('self');
+            $method3->hasReturnType()->willReturn(true);
+            $method3->getReturnType()->willReturn('parent');
         }
 
         $classNode   = $this->reflect($class, array());
@@ -67,8 +77,14 @@ class ClassMirrorSpec extends ObjectBehavior
         $methodNodes->shouldHaveCount(3);
 
         $classNode->hasMethod('getName')->shouldReturn(true);
-        $classNode->hasMethod('isPublic')->shouldReturn(true);
-        $classNode->hasMethod('isAbstract')->shouldReturn(true);
+        $classNode->hasMethod('getSelf')->shouldReturn(true);
+        $classNode->hasMethod('getParent')->shouldReturn(true);
+
+        if (version_compare(PHP_VERSION, '7.0', '>=')) {
+            $classNode->getMethod('getName')->getReturnType()->shouldReturn('string');
+            $classNode->getMethod('getSelf')->getReturnType()->shouldReturn('\Custom\ClassName');
+            $classNode->getMethod('getParent')->getReturnType()->shouldReturn('\Custom\ParentClassName');
+        }
     }
 
     /**
--- a/src/Prophecy/Doubler/Generator/ClassMirror.php
+++ b/src/Prophecy/Doubler/Generator/ClassMirror.php
@@ -143,7 +143,17 @@ class ClassMirror
         }
 
         if (version_compare(PHP_VERSION, '7.0', '>=') && true === $method->hasReturnType()) {
-            $node->setReturnType((string) $method->getReturnType());
+            $returnType = (string) $method->getReturnType();
+            $returnTypeLower = strtolower($returnType);
+
+            if ('self' === $returnTypeLower) {
+                $returnType = $method->getDeclaringClass()->getName();
+            }
+            if ('parent' === $returnTypeLower) {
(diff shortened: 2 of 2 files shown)


src/Prophecy/Doubler/Generator/ClassMirror.php
```diff
@@ -143,7 +143,9 @@ private function reflectMethodToNode(ReflectionMethod $method, Node\ClassNode $c
         }
 
         if (version_compare(PHP_VERSION, '7.0', '>=') && true === $method->hasReturnType()) {
-            $node->setReturnType((string) $method->getReturnType());
+            $returnType = (string) $method->getReturnType();
+            $returnType = 'self' === $returnType ? $classNode->getParentClass() : $returnType;
```
aik099: To what class `self` is resolved in PHP 7?

If it's the same logic as with `@return self` in DocBlocks, then it's class, where such DocBlock was used, not necessarily parent class.

aik099: According to https://wiki.php.net/rfc/return_types RFC two special return types are allowed:

* `self`
* `parent`

The RFC however doesn't explain further what `self` means. It's just a theory, but since we have a `parent` return type, then maybe `self` acts as `static` (means this class) in DocBlocks.

ciaranmcnulty: It's the same logic as PHP; `self` means 'the class in which the declaration is found`.

aik099:
You're saying yourself that it doesn't work in pure PHP. How Prophecy can help with that.

ciaranmcnulty:
@aik099 My reading is that Prophecy is currently making the double return `self` which causes an error, and this patch makes it return the parent type?

src/Prophecy/Doubler/Generator/ClassMirror.php
aik099: But the following code (same idea as in code from PR description) fails on PHP7 with `Fatal error: Declaration of A2::test(): A2 must be compatible with A1::test(): A1 in /path/to/file.php on line ...
` error:

```php
class A1 {

	function test(): self
	{
		return new static();
	}
}

class A2 extends A1 {

	function test(): self
	{
		return new static();
	}
}

$a3 = new A2();
$a3->test();
```

aik099:
> @aik099 My reading is that Prophecy is currently making the double return self which causes an error, and this patch makes it return the parent type?

That's what I've understood from code as well. But I think it's incorrect, because `parent class` isn't what `self` refers to. It should be `this class`.

aik099:
The problem, that is solved in fact is that `: self` in mocked method is replaced with `: \self` in mock and that is what is fixed in the first place.

bendavies:
indeed, that's correct.

aik099:
@bendavies , why just not keep `self` and `parent` keywords as-is without trying to resolve them to any actual class?

bendavies:
@aik099 that would result in the Compile Error as in the PR description.

aik099:
Ah, now I see why you return parent class. It's because parent class is mocked and actual `self` reffers to mock sub-class :smile:

aik099:
@bendavies , maybe add a comment in there to what "parent class" refers to and why it's not current class name.

bendavies:
I've updated the code, it wasn't correct.
The return type is now the class where the method is defined.

aik099:
We should add some PHP7-only test that confirm, that `self` return type is properly handled.

bendavies:
will do

bendavies:
@ciaranmcnulty @aik099 updated with parent support, and tests.

spec/Prophecy/Doubler/Generator/ClassMirrorSpec.php
```diff
@@ -67,8 +77,14 @@ function it_reflects_a_class_by_mirroring_all_its_public_methods(
...
 
         $classNode->hasMethod('getName')->shouldReturn(true);
-        $classNode->hasMethod('isPublic')->shouldReturn(true);
-        $classNode->hasMethod('isAbstract')->shouldReturn(true);
+        $classNode->hasMethod('getSelf')->shouldReturn(true);
+        $classNode->hasMethod('getParent')->shouldReturn(true);
+
+        if (version_compare(PHP_VERSION, '7.0', '>=')) {
+            $classNode->getMethod('getName')->getReturnType()->shouldReturn('string');
+            $classNode->getMethod('getSelf')->getReturnType()->shouldReturn('\Custom\ClassName');
```
aik099: Are you sure, that `getReturnType` will return FCQN with leading `\`?

src/Prophecy/Doubler/Generator/ClassMirror.php
```diff
@@ -143,7 +143,17 @@ private function reflectMethodToNode(ReflectionMethod $method, Node\ClassNode $c
...
 
         if (version_compare(PHP_VERSION, '7.0', '>=') && true === $method->hasReturnType()) {
-            $node->setReturnType((string) $method->getReturnType());
+            $returnType = (string) $method->getReturnType();
+            $returnTypeLower = strtolower($returnType);
+
+            if ('self' === $returnTypeLower) {
+                $returnType = $method->getDeclaringClass()->getName();
+            }
+            if ('parent' === $returnTypeLower) {
```
aik099: I think `elseif` (or `else if`, not sure what coding standard says) would work better in this case.

src/Prophecy/Doubler/Generator/ClassMirror.php
```diff
@@ -143,7 +143,17 @@ private function reflectMethodToNode(ReflectionMethod $method, Node\ClassNode $c
...
+                $returnType = $method->getDeclaringClass()->getParentClass()->getName();
```
aik099: I wasn't able to find any documentation about return type `parent`. Are you certain that `parent` means what you've implemented?

spec/Prophecy/Doubler/Generator/ClassMirrorSpec.php
bendavies: I think you are misreading this. This is an expectation on the returned MethodNode.

src/Prophecy/Doubler/Generator/ClassMirror.php
bendavies: I'm not sure why else if is better here, but if there is a coding standard I'll obviously follow.

spec/Prophecy/Doubler/Generator/ClassMirrorSpec.php
aik099: No, this is the value that you return when this method is called. If this would be real (not test) code, then `getReturnType` would return `\Custom\...` or `Custom\...`?

src/Prophecy/Doubler/Generator/ClassMirror.php
bendavies: Reasonably. http://php.net/manual/en/keyword.parent.php
This is also how doctrine common implementation it.

aik099: It doesn't explicitly say what `parent` means when used a return type hint, but we'll assume that it behaves the same in that case as well.

spec/Prophecy/Doubler/Generator/ClassMirrorSpec.php
bendavies: I think you're saying that i'm creating a return promise here, but I'm not.
`shouldReturn` creates an expection and checks that `getReturnType` actually returns `\Custom\ClassName`

You can check yourself. Change the `\Custom\ClassName` value and run the tests.

src/Prophecy/Doubler/Generator/ClassMirror.php
aik099: I don't know if there is a standard, but if both conditions obviously can't happen together using `elseif` would also indicate that in the code.

spec/Prophecy/Doubler/Generator/ClassMirrorSpec.php
aik099: OK.

bendavies:
@ciaranmcnulty is this mergable?

ciaranmcnulty:
Looks good to me. I don't normally merge things in Prophecy

bendavies:
no problem, who can i ping then?

aik099:
Who do the merges. Please add me as maintainer so that I can merge as well.

stof:
Please also support them in typehints (relevant for all versions), not only in return types

bendavies:
@stof good point, thanks.

aik099:
> Please also support them in typehints (relevant for all versions), not only in return types

You're saying that even in PHP5 the following code can be valid:

```php

class A extends B {

    public function test(parent $a, self $b)
    {
        ...
    }
}
```

At least this was my understanding of `relevant for all versions` phrase.

stof:
@aik099 yes, it is valid (at least as of 5.3, not sure about older 5.x versions but we don't care)

everzet:

aik099:
I guess @bendavies should send another PR now, where he adds `self` and `parent` type hint support for method arguments.

bendavies:
@aik099 if you're like to contribute, go ahead.

aik099:
I've filed an issue at #256 as a reminder, that this needs to be implemented.

bendavies:
If you have time, go for it

PR #255, opened 2016-02-15
Use 7.0 rather than 7

GrahamCampbell:


aik099:
The change would prevent PHP 7.1 to be used instead of PHP 7.0, when PHP 7.1 will be released I presume?

GrahamCampbell:
It does prevent 7.1, on purpose. You still want to test on 7.0 when 7.1 comes out.

Jean85:
:+1:

everzet:

PR #260, opened 2016-02-18
Ignore empty methods from phpdoc to avoid syntax erros in generated code

ftrrtf:
When the methods are described incorrectly in phpdoc `ClassCodeGenerator` creates incorrect code with a syntax error:

```
....
public  function () {
}
...
```
`PHP Parse error:  syntax error, unexpected '(', expecting identifier (T_STRING)`


everzet:
I wish I never accepted support for magic methods into the core. Oh well...

everzet:

Issue #250, opened 2016-01-28
Spies and PHP7 return typehinting


Sam-Burns:
Spies don't work with PHP return typehinting.

Example in repo at https://github.com/Sam-Burns/prophecy-bug-demo for now ; code cut-and-paste below for prosperity.

Reproducible step: Run PHPSpec.
Expected result: Spec passes.
Actual result:
```
✘ Fatal error happened while executing the following
    it calls something on its dependency
    Uncaught TypeError: Return value of Double\DependencyWithReturnTypehinting\P1::getAnInt() must be of the type integer, null returned in
```

The basic problem seems to be that spies don't return anything, so you can't make a spy of a class which uses return typehinting.  Perhaps a solution might be to catch ```\Throwable```, ```\Error``` or something when calling methods on spies, but I don't really know because I've never read the Prophecy codebase.

The code in the temporary repo demonstrating the bug is (was) as follows:

```php
<?php
class SubjectUnderTest
{
    public function getAnIntFromThisThing(DependencyWithReturnTypehinting $dependencyWithReturnTypehinting)
    {
        $int = $dependencyWithReturnTypehinting->getAnInt();
    }
}
```

```php
<?php
class DependencyWithReturnTypehinting
{
    public function getAnInt(): int
    {
        return 42;
    }
}
```

```php
<?php
namespace spec;
use PhpSpec\ObjectBehavior;
class SubjectUnderTestSpec extends ObjectBehavior
{
    function it_calls_something_on_its_dependency(\DependencyWithReturnTypehinting $dependencyWithReturnTypehinting)
    {
        $this->getAnIntFromThisThing($dependencyWithReturnTypehinting);
        $dependencyWithReturnTypehinting->getAnInt()->shouldHaveBeenCalled();
    }
}
```

Jean85:
Confirmed also under Prophecy 1.6.0, using PHP 7.0.2

Sam-Burns:
@ciaranmcnulty Further to discussion at #phpuk16, the bug report is still valid for Prophecy 1.6.0.  I've updated the repo with the reproducible example for reference.

everzet:
@Sam-Burns care to PR the failing test cases to Prophecy? That will make it easier to fix.

Sam-Burns:
@everzet Will be a bit of a messy pull request, but will do.

Sam-Burns:
[...]

PR #261, opened 2016-02-22
Capturing bug with PHP7 return typehinting and spies

Sam-Burns:
Capturing bug with integration test, pending clearer idea of which class/unit test is relevant


Sam-Burns:
Ref https://github.com/phpspec/prophecy/issues/250

Sam-Burns:
I really don't think you will want to merge this pull request, but @everzet was asking for a test capturing the bug in the Prophecy repo.  Any guidance as to which class is responsible for the bug, and which spec this can be moved to, would be very much appreciated.

aik099:
Can you please change your test to use PHPSpec instead of PHPUnit?

Sam-Burns:
@aik099 No, because it's not clear to me what class the bug is in.  Do you know the answer to that?

aik099:
No.

Sam-Burns:
@aik099 On the original bug report (https://github.com/phpspec/prophecy/issues/250) @everzet asked me to put a test in the repo.  I suggest we leave it as it is for now, until someone comes along who knows which spec this can be transferred to.  Then we can get rid of the PHPUnit installation and the integration test, and capture the buggy behaviour in a unit test instead.

Alternatively, I could use Behat to write an integration test, but I'm not sure if that improves the situation at all.

everzet:
@Sam-Burns this is perfect, Sam. Working on it

everzet:
Ok. I don't see a problem here. In my view, Prophecy acts exactly like it should.

You double `CollaboratorWithReturnTypehinting::getAnInt(): int`. Double calls do nothing - return `null` by default. If you do not explicitly tell it to return a particular integer, it would break the type hint, yes.

Doing logical here:

```php
    $subject->getAnInt()->willReturn(42);
    $this->getAnIntFromThisThing($subject);
    $subject->getAnInt()->shouldHaveBeenCalled();
```

fixes the fatal. As it should.

ciaranmcnulty:
This is inconsistent, IIRC prophecy is returning non-null object instances where return type hints are used

everzet:
@ciaranmcnulty I am not aware of such behaviour.

jakzal:
> prophecy is returning non-null object instances where return type hints are used

What do you think of implementing such a behaviour? I mean returning dummies of objects and "empty" values for simple types (0, '', false).

PR #263, opened 2016-03-01
Upgrade to reflection-docblock 3

dunglas:
Tests will be green when phpDocumentor/TypeResolver#16 and phpDocumentor/ReflectionDocBlock#72 will be merged.

### DIFF
--- a/src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
+++ b/src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
@@ -59,14 +59,11 @@ class MagicCallPatch implements ClassPatchInterface
 
         try {
             $phpdoc = $this->docBlockFactory->create($reflectionClass, $this->contextFactory->createFromReflector($reflectionClass));
+            $tagList = $phpdoc->getTagsByName('method');
         } catch (\InvalidArgumentException $e) {
             // No DocBlock
+            $tagList = array();
         }
-
-        /**
-         * @var Method[] $tagList
-         */
-        $tagList = isset($phpdoc) ? $phpdoc->getTagsByName('method') : array();
 
         $interfaces = $reflectionClass->getInterfaces();
         foreach($interfaces as $interface) {
@@ -80,6 +77,10 @@ class MagicCallPatch implements ClassPatchInterface
 
         foreach($tagList as $tag) {
             $methodName = $tag->getMethodName();
+
+            if (empty($methodName)) {
+                continue;
+            }
 
             if (!$reflectionClass->hasMethod($methodName)) {
                 $methodNode = new MethodNode($tag->getMethodName());
--- a/src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
+++ b/src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
@@ -11,7 +11,10 @@
 
 namespace Prophecy\Doubler\ClassPatch;
 
-use phpDocumentor\Reflection\DocBlock;
+use phpDocumentor\Reflection\DocBlock\Tags\Method;
+use phpDocumentor\Reflection\DocBlockFactory;
+use phpDocumentor\Reflection\DocBlockFactoryInterface;
+use phpDocumentor\Reflection\Types\ContextFactory;
 use Prophecy\Doubler\Generator\Node\ClassNode;
 use Prophecy\Doubler\Generator\Node\MethodNode;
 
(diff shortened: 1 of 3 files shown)


src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
```diff
@@ -44,23 +57,30 @@ public function apply(ClassNode $node)
...
+            $phpdoc = $this->docBlockFactory->create($reflectionClass, $this->contextFactory->createFromReflector($reflectionClass));
+        } catch (\InvalidArgumentException $e) {
+            // No DocBlock
+        }
 
-        $tagList = $phpdoc->getTagsByName('method');
+        /**
+         * @var Method[] $tagList
+         */
+        $tagList = isset($phpdoc) ? $phpdoc->getTagsByName('method') : array();
```
stof: I don't like the idea of dealing with codepath where the variable may be defined or no. The code should probably be refactored by moving this inside the try/catch

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
```diff
@@ -44,23 +57,30 @@ public function apply(ClassNode $node)
...
-        $phpdoc = new DocBlock($reflectionClass->getDocComment());
+        try {
+            $phpdoc = $this->docBlockFactory->create($reflectionClass, $this->contextFactory->createFromReflector($reflectionClass));
+        } catch (\InvalidArgumentException $e) {
+            // No DocBlock
+        }
 
-        $tagList = $phpdoc->getTagsByName('method');
+        /**
+         * @var Method[] $tagList
```
sroze: If you really want it, I would use a single line comment instead.

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
```diff
@@ -19,9 +22,19 @@
...
  *
  * @author Thomas Tourlourat <thomas@tourlourat.com>
+ * @author Kévin Dunglas <dunglas@gmail.com>
  */
 class MagicCallPatch implements ClassPatchInterface
 {
+    private $docBlockFactory;
+    private $contextFactory;
+
+    public function __construct()
```
sroze: Can you make these dependencies as optionally injectable?

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
```diff
@@ -44,23 +57,30 @@ public function apply(ClassNode $node)
...
+                $tagList = array_merge($tagList, $phpdoc->getTagsByName('method'));
+            } catch (\InvalidArgumentException $e) {
+                // No DocBlock
+            }
         }
 
         foreach($tagList as $tag) {
             $methodName = $tag->getMethodName();
 
-            if (empty($methodName)) {
```
sroze: You're sure that with the version 3.0, it won't return empty method names? You're reverting a 12 days old PR (#260) here.

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
stof: I don't think it is worth it. You generally don't instantiate MagicCallPatch yourselves anyway.

And exposing them would mean that we then have to provide backward compatibility with people injecting us a v3 DocBlockFactory.
We can still change this later if there is a use case to customize the parsing of docblocks

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
stof: Well, the testsuite will tell us

stof:
This PR means that the min requirement of Prophecy is now PHP 5.5 though

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
sroze: Yep, this is fair enough we can open the doors after.

aik099:
So Symfony wants Reflection v3, but Prophecy wants Reflection v2?

sroze:
We need to be able to support the two versions, else it will again cause a lot of problems. Based on the small amount of changes that are required, that's fair to abstract it and uses a basic `class_exists` to create the correct adapter here IMO.

dunglas:
I fully agree, will do it ASAP.

dunglas:
Closing in favor of #264

PR #265, opened 2016-03-27
Allow Returning by reference

pharazon:
I'm building mocks for an old version of ADODB database abstraction library, where the main $db->Execute() returns a reference to the result set. Without this code, my mocks works ok, so I don't know what is the reason why returning by reference is not allowed?

If there is a case where not allowing it is useful, maybe there could be a configurable parameter for that. I vote for allowing it by default.


aik099:
Maybe it's considered a bad practice and that's why it'd disallowed.

stof:
This won't work, as the double would still not be able to return a reference, as none of the Prophecy internals are working with references (and this won't change)

pharazon:
We'll then I need to use this fork because mainline Prophecy doesn't work with AdoDB.

PR #269, opened 2016-04-29
Removed excess docblock

OskarD:
I accidentally found this while searching my project for the same docblock after noticing it and removing it from my PhpStorm template. I thought you probably don't want it either :)


stof:
thanks

PR #274, opened 2016-05-30
Convert the ClassMirror spec to an integration test

stof:
Mocking the Reflection API does not ensure things work fine on all PHP versions (as we might not reflect the actual behavior) and is much more complex than writing an integration test.

Thus, this respects the best practice of not mocking what we don't own.

I have not yet reviewed the full spec suite to identify other classes which should use an integration test rather than specs due to crossing boundaries, but this can be done in separate PRs anyway now that the structure is in place.
We should probably also add some functional tests of the library (other than the fact that phpspec uses Prophecy and so some kind of failures would break the testsuite)

TODOs:
- [x] update contributing instructions to mention PHPUnit
- [x] skip the test for typehinted variadic argument on HHVM (feature not supported)

Note that you need to run PHPUnit from the composer-installed version, not from the phar. Due to the way PHPUnit autoloads its classes when using the PHAR, you would be running the Prophecy code embedded in the Phar rather than the local codebase if you use the Phar.


ciaranmcnulty:
This looks good - I feel that the tests could perhaps be split into multiple TestCases?

stof:
Well, I follow the convention of having 1 test case per tested class (this is a test covering 1 class, but a class at the domain boundary)

PR #276, opened 2016-06-06
Bumped branch alias

GrahamCampbell:
I assume the next version will be 1.7.0?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
 
     "extra": {
         "branch-alias": {
-            "dev-master": "1.6.x-dev"
+            "dev-master": "1.7.x-dev"
         }
     }
 }


ptondereau:
Will this fix symfony 3.1.* constraint bug with phpunit  prophecy?

GrahamCampbell:
Take that as a no. :P

stof:
I don't know whether it will be 1.7.0 or 1.6.2

GrahamCampbell:
I was referring to the 1.6.1 release you just did. ;)

GrahamCampbell:
For now, we should probably assume the next release will be 1.6.2, unless you have a reason to make it 1.7, then we can change it only then, right?

stof:
exactly

PR #278, opened 2016-06-14
Fix incorrect ObjectStateToken method argument order

Anyqax:
this doesn't actually cause a visible misbehaviour, but is obviously still technically wrong;
since there is no functional change, no spec/test possible


everzet:
👍

stof:
This broke the ability to use a prophecy object as value in ``Argument::which``

PR #285, opened 2016-09-14
Add PHP 7.1 to Travis

danizord:
I did some tests with PHP 7.1 and Prophecy locally and it seems like Prophecy does not work with PHP 7.1, so I'm adding it to the build matrix to see if tests passes.


danizord:
Test passes, but I've found an issue with optional arguments, reported in #286

jakzal:
:+1:

everzet:

PR #288, opened 2016-09-26
Allow \Error instances in ThrowPromise

### LINKED ISSUE #281: willThrow does not support throwing Errors on PHP7
As the title say, ther is no possibility to mock throwing Errors in PHP7 using prophecy.

```
$prophecyMethod->willThrow(new Error());
```

This should be possible but is not.

jameshalsall:
Fixes #281.

The build will fail until the corresponding PHPSpec PR (https://github.com/phpspec/phpspec/pull/1006) is merged into a stable release.

### DIFF
--- a/src/Prophecy/Promise/ThrowPromise.php
+++ b/src/Prophecy/Promise/ThrowPromise.php
@@ -34,7 +34,7 @@ class ThrowPromise implements PromiseInterface
     /**
      * Initializes promise.
      *
-     * @param string|\Exception $exception Exception class name or instance
+     * @param string|\Exception|\Throwable $exception Exception class name or instance
      *
      * @throws \Prophecy\Exception\InvalidArgumentException
      */
--- a/src/Prophecy/Promise/ThrowPromise.php
+++ b/src/Prophecy/Promise/ThrowPromise.php
@@ -94,6 +94,6 @@ class ThrowPromise implements PromiseInterface
      */
     private function isAValidThrowable($exception)
     {
-        return is_a($exception, 'Exception') || is_subclass_of($exception, 'Throwable');
+        return is_a($exception, 'Exception', true) || is_subclass_of($exception, 'Throwable', true);
     }
 }
--- a/src/Prophecy/Promise/ThrowPromise.php
+++ b/src/Prophecy/Promise/ThrowPromise.php
@@ -41,7 +41,7 @@ class ThrowPromise implements PromiseInterface
     public function __construct($exception)
     {
         if (is_string($exception)) {
-            if (!class_exists($exception) || $this->isNotAValidThrowable($exception)) {
+            if (!class_exists($exception) || !$this->isAValidThrowable($exception)) {
                 throw new InvalidArgumentException(sprintf(
                     'Exception / Throwable class or instance expected as argument to ThrowPromise, but got %s.',
                     $exception
@@ -87,8 +87,13 @@ class ThrowPromise implements PromiseInterface
         throw $this->exception;
     }
 
-    private function isNotAValidThrowable($exception)
+    /**
+     * @param string $exception
+     *
+     * @return bool
+     */
+    private function isAValidThrowable($exception)
     {
-        return 'Exception' !== ltrim($exception, '\\') && !is_subclass_of($exception, 'Exception') && !is_subclass_of($exception, 'Throwable');
+        return is_a($exception, 'Exception') || is_subclass_of($exception, 'Throwable');
     }
 }
--- a/src/Prophecy/Promise/ThrowPromise.php
+++ b/src/Prophecy/Promise/ThrowPromise.php
@@ -41,17 +41,15 @@ class ThrowPromise implements PromiseInterface
     public function __construct($exception)
     {
         if (is_string($exception)) {
-            if (!class_exists($exception)
-             && 'Exception' !== $exception
-             && !is_subclass_of($exception, 'Exception')) {
+            if (!class_exists($exception) || $this->isNotAValidThrowable($exception)) {
                 throw new InvalidArgumentException(sprintf(
-                    'Exception class or instance expected as argument to ThrowPromise, but got %s.',
+                    'Exception / Throwable class or instance expected as argument to ThrowPromise, but got %s.',
                     $exception
                 ));
             }
         } elseif (!$exception instanceof \Exception && !$exception instanceof \Throwable) {
             throw new InvalidArgumentException(sprintf(
-                'Exception/Error class or instance expected as argument to ThrowPromise, but got %s.',
+                'Exception / Throwable class or instance expected as argument to ThrowPromise, but got %s.',
                 is_object($exception) ? get_class($exception) : gettype($exception)
             ));
         }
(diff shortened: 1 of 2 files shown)


src/Prophecy/Promise/ThrowPromise.php
```diff
@@ -49,9 +49,9 @@ public function __construct($exception)
                     $exception
                 ));
             }
-        } elseif (!$exception instanceof \Exception) {
+        } elseif (!$exception instanceof \Exception && !$exception instanceof \Error) {
```
ciaranmcnulty: Probably better to check against Throwable:
`!$exception instanceof \Exception && !$exception instanceof \Throwable`

This futureproofs against future Throwables

jameshalsall: Updated

ciaranmcnulty:
This looks good to me

stof:
This is missing the case of class names passed to the ThrowPromise. You still only allow exceptions in this case

jameshalsall:
There's no example that covers it, but it already works I believe (see https://github.com/phpspec/prophecy/pull/288/files#diff-00571e066472de6498928505cee554f5R43) - I will add an example later on today.

stof:
no, because inside the case of is_string, there is a check whether the string is a class name being a subclass of Exception. It does not allow Error

ciaranmcnulty:
Good spot, @stof

jameshalsall:
The condition inside the is_string condition looks like it would be fine (PHPSpec also has a test for passing `'\Error'` to the ThrowMatcher (which uses the ThrowPromise), and that passes.

```
if (!class_exists($exception)
              && 'Exception' !== $exception
              && !is_subclass_of($exception, 'Exception')) {
```

From what I can see only if the class doesn't exist would it go on to throw an exception

ciaranmcnulty:
@jameshalsall The desired behaviour is that it throws the `$exception` so you can do:

`$this->foo()->shouldThrow('\Error')`

The string needs to be checked as a Throwable.

jameshalsall:
I don't think that's how it's currently working, if you look at the constructor logic carefully it will not throw an exception when you pass `"\Error"` as the argument. This is a class that exists, so the `!class_exists()` check would fail and skip the other conditions.

stof:
Well, it means that there is a bug currently.

ciaranmcnulty:
Ah, probably the `&&` should be `||`. Looks like we're trying to throw anything we're given, even if it's not throwable.

jameshalsall:
I've updated this with more examples to cover the broken logic around the ThrowPromise accepting any class rather than only throwables, and the logic has been fixed too.

This might constitute a BC break, so let me know if you want it moving to another PR.

ciaranmcnulty:
IMO it's not a BC break - It's clearly a bug and anyone passing a type that isn't `Throwable` will be getting a later error when Prophecy tries to throw it.

src/Prophecy/Promise/ThrowPromise.php
```diff
@@ -88,4 +86,9 @@ public function execute(array $args, ObjectProphecy $object, MethodProphecy $met
 
         throw $this->exception;
     }
+
+    private function isNotAValidThrowable($exception)
+    {
+        return 'Exception' !== ltrim($exception, '\\') && !is_subclass_of($exception, 'Exception') && !is_subclass_of($exception, 'Throwable');
```
ciaranmcnulty: Rather than `ltrim`, which feels hacky, we could use `instanceof` here.

ciaranmcnulty: We could use `instanceof` here rather than messing around with `ltrim`

jameshalsall: Yeah you're right. Not sure why I did it like that.... I'm away for the next few days but will fix it next week.

ciaranmcnulty: It was like that previously before you fixed the booleans. Doesn't block a merge.

everzet:
@jameshalsall apply final changes & I'll merge it.

jameshalsall:
Are there any outstanding changes to be made on this?

src/Prophecy/Promise/ThrowPromise.php
stof: @ciaranmcnulty you cannot, as a class name is not an instance of the class. It is a string

jameshalsall:
@everzet @ciaranmcnulty is there anything preventing this from being merged?

src/Prophecy/Promise/ThrowPromise.php
```diff
@@ -41,17 +41,15 @@ class ThrowPromise implements PromiseInterface
...
-             && !is_subclass_of($exception, 'Exception')) {
+            if (!class_exists($exception) || !$this->isAValidThrowable($exception)) {
                 throw new InvalidArgumentException(sprintf(
-                    'Exception class or instance expected as argument to ThrowPromise, but got %s.',
+                    'Exception / Throwable class or instance expected as argument to ThrowPromise, but got %s.',
                     $exception
                 ));
             }
-        } elseif (!$exception instanceof \Exception) {
+        } elseif (!$exception instanceof \Exception && !$exception instanceof \Throwable) {
```
avant1: If this class constructor allows any `\Throwable` as parameter its PHPDoc could be updated to reflect this.
And the same for `MethodProphecy::willThrow()`.
Does PHPDoc allows 3 or more parameter types options like `string|\Exception|\Throwable`?

stof: @avant1 your comment is wrong: https://3v4l.org/LN2Xk

avant1: Haha, i've made the following test before commenting: https://3v4l.org/kT0MT
And I was not careful while reading error message. My bad.

ciaranmcnulty: It's only metadata :-)

jameshalsall: Prophecy still supports PHP 5.3, so I'll go with `string|\Exception|\Throwable`

jameshalsall: @avant1 updated

unfunco: I'd put `\Throwable` before `\Exception` since `\Exception` inherits from `\Throwable`, you'd reach the condition earlier on PHP `>=7.1` – minor quibble.

src/Prophecy/Promise/ThrowPromise.php
```diff
@@ -88,4 +86,14 @@ public function execute(array $args, ObjectProphecy $object, MethodProphecy $met
...
     }
+
+    /**
+     * @param string $exception
+     *
+     * @return bool
+     */
+    private function isAValidThrowable($exception)
+    {
+        return is_a($exception, 'Exception', true) || is_subclass_of($exception, 'Throwable', true);
```
unfunco: Same here, I think the logical or conditions should be reversed.

everzet:
@jameshalsall again, thanks for hard work. Much-much appreciated. And sorry for a long delay

PR #290, opened 2016-09-28
Add support for detecting @method on interfaces that the class itself implements, or when the stubbed class is an interface itself

### LINKED ISSUE #284: Mocked @method throws MethodNotFoundException.
I've seen that the following has been merged #247 however I am having trouble getting it to work.

My test:
```
    /**
     * @test
     */
    public function it_will_returned_cached_version_if_set()
    {
        $redis = $this->prophesize(Client::class);
        $next = $this->prophesize(DealFinder::class);

        $expectedDealId = '6979737c-2bd0-4988-aacd-195668e3703a';
        $expectedResponse = ['deal_id' => $expectedDealId];
        
        $redis->exists(Argument::containing($expectedDealId))->willReturn(true);
        $redis->get(Argument::containing($expectedDealId))->willReturn(json_encode($expectedResponse));

        $SUT = new CachedDealFinder($redis->reveal(), $next->reveal());
        $result = $SUT->findWithId($expectedDealId);

        static::assertEquals($expectedResponse, $result);
    }
```

Produces:
```
1) CachedDealFinderTest::it_will_returned_cached_version_if_set
Prophecy\Exception\Doubler\MethodNotFoundException: Method `Double\Predis\Client\P1::exists()` is not defined.
```

Version:
```
composer info -i | grep prophecy
phpspec/prophecy v1.6.1 Highly opinionated mocking framework for PHP 5.3+
```

Seldaek:
Fixes #284 (well kinda, it also relies on a bug being fixed in phpdocumentor, but it fixes the prophecy part.. link to other PR coming soon)

### DIFF
--- a/src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
+++ b/src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
@@ -51,23 +51,27 @@ class MagicCallPatch implements ClassPatchInterface
      */
     public function apply(ClassNode $node)
     {
-        $parentClass = $node->getParentClass();
-        $reflectionClass = new \ReflectionClass($parentClass);
+        $types = array_filter($node->getInterfaces(), function ($interface) {
+            return 0 !== strpos($interface, 'Prophecy\\');
+        });
+        $types[] = $node->getParentClass();
 
-        $tagList = $this->tagRetriever->getTagList($reflectionClass);
+        foreach ($types as $type) {
+            $reflectionClass = new \ReflectionClass($type);
+            $tagList = $this->tagRetriever->getTagList($reflectionClass);
 
-        foreach($tagList as $tag) {
-            $methodName = $tag->getMethodName();
+            foreach($tagList as $tag) {
+                $methodName = $tag->getMethodName();
 
-            if (empty($methodName)) {
-                continue;
-            }
+                if (empty($methodName)) {
+                    continue;
+                }
 
-            if (!$reflectionClass->hasMethod($methodName)) {
-                $methodNode = new MethodNode($methodName);
-                $methodNode->setStatic($tag->isStatic());
-
-                $node->addMethod($methodNode);
+                if (!$reflectionClass->hasMethod($methodName)) {
+                    $methodNode = new MethodNode($methodName);
+                    $methodNode->setStatic($tag->isStatic());
+                    $node->addMethod($methodNode);
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


stof:
can you give a link to the phpdocumentor bug ?

Seldaek:
Patience, as I said I was working on the fix :) Here it is https://github.com/phpDocumentor/ReflectionDocBlock/pull/84

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
```diff
@@ -51,23 +51,27 @@ public function supports(ClassNode $node)
...
     {
-        $parentClass = $node->getParentClass();
-        $reflectionClass = new \ReflectionClass($parentClass);
+        $classNames = array_filter($node->getInterfaces(), function ($interface) {
+            return 0 !== strpos($interface, 'Prophecy\\');
+        });
+        $classNames[] = $node->getParentClass();
 
-        $tagList = $this->tagRetriever->getTagList($reflectionClass);
+        foreach ($classNames as $interface) {
```
ciaranmcnulty: Naming is weird here - `$classNames` contains a mix of classes and interfaces. Maybe `$types as $type`?

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
```diff
@@ -51,23 +51,27 @@ public function supports(ClassNode $node)
      */
     public function apply(ClassNode $node)
     {
-        $parentClass = $node->getParentClass();
-        $reflectionClass = new \ReflectionClass($parentClass);
+        $classNames = array_filter($node->getInterfaces(), function ($interface) {
```
ciaranmcnulty: This only returns the interfaces that the class has been told to double - to handle possible annotations on parents you might need to call `class_implements` instead.

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
```diff
@@ -51,23 +51,27 @@ public function supports(ClassNode $node)
...
+            return 0 !== strpos($interface, 'Prophecy\\');
+        });
+        $classNames[] = $node->getParentClass();
```
ciaranmcnulty: Similarly to my other comment, this might need to call `class_parents` rather than asking what class we have explicitly doubled.

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
Seldaek: I am not sure, I'll tweak the test to verify but I believe PHPDoc might do the right thing and traverse parents itself. Other stuff makes sense too, I'll do this later :)

ciaranmcnulty: Ah! If that's the case maybe a test would help clarify it

Seldaek:
Alright added some test and fixed naming, it seems fine to me now.

everzet:
@Seldaek cheers Jordi!

PR #293, opened 2016-11-07
Allow __soapCall to be mocked.

frankdejonge:
PHP's SoapClient has a public method `__soapCall` which is basically `__call` with a special signature. This change allows this method to be mocked.

### DIFF
--- a/fixtures/SpecialMethods.php
+++ b/fixtures/SpecialMethods.php
@@ -33,4 +33,8 @@ class SpecialMethods
     {
     }
 
+    function __soapCall($function_name, array $arguments, array $options = null, $input_headers = null, array &$output_headers = null)
+    {
+    }
+
 }
--- a/src/Prophecy/Doubler/Generator/ClassMirror.php
+++ b/src/Prophecy/Doubler/Generator/ClassMirror.php
@@ -32,6 +32,7 @@ class ClassMirror
         '__wakeup',
         '__toString',
         '__call',
+        '__soapCall',
         '__invoke'
     );
 
--- a/tests/Doubler/Generator/ClassMirrorTest.php
+++ b/tests/Doubler/Generator/ClassMirrorTest.php
@@ -17,7 +17,7 @@ class ClassMirrorTest extends \PHPUnit_Framework_TestCase
 
         $node = $mirror->reflect($class, array());
 
-        $this->assertCount(7, $node->getMethods());
+        $this->assertCount(8, $node->getMethods());
     }
 
     /**


stof:
Next question is why you are mocking SoapClient. It is not your own class, so can you guarantee you know how it behaves for real ? If the answer is no, mocking the class is a lie, as you cannot ensure your mock behaves the right way, and so your test does not ensure that your code is actually working.

frankdejonge:
@stof I'm well aware of that. I'm surely not the only one who knowingly doesn't make the "best" call, but a good enough one. While I generally agree with your stance, it's not "following rules set in stone" that makes something good. It's knowing when not to make the bad calls.

Taluu:
Sometimes it is not "mocking", but "stubbing" that is needed (preventing a call to whatever external api we shouldn't care about), e.g "I know it makes this call, and I don't care about the result" (or "it should return this, I don't care how", but then you need to update once in a while if the output changes, indeed).

But instead of adding a new method (`__soapCall`), how about removing the fact that `_` is not considered as "mockable" ?

frankdejonge:
I'd be totally fine with that too.

frankdejonge:
I've added an alternative PR which does just that.

jakzal:
@Taluu in this context "mocking" was used colloquially. 3rd party classes should not be test doubled in general.

frankdejonge:
@jakzal in general that's true, however there are exceptions to every rule, as is the case with this one.

jakzal:
I don't see why you'd make an exception in this case, especially when it's easier to introduce your own abstraction rather than rely on a class with so much magic inside.

frankdejonge:
In some cases the party where you want to interact with might not be reliable enough to be a determining part of your test suite, or they may not provide a test or acceptance environment. In this case you, as a developer, can do two things. Either mock out a client, such as the SoapClient, or create your own soap interface to stub the service. In this case the costs of creating a stubbed implementation simply outweigh the common good practise to "not mock that you don't own". As such, "real life" arguments sometimes outweigh theoretically sound guidelines. The question now becomes, should we enforce the guidelines without regard of how situation actually play out. Or do we acknowledge a developer is responsible for making such calls and do we lift the restriction we placed upon them. I personally think the second option is a much more realistic approach.

stof:
@frankdejonge but if you mock the external code, your testsuite provides you **exactly no guarantee** that your code works fine. It tells you that it works fine *if your own understanding of the external code was right and is still right*.
This means that it does not tell you that it still works today (the other system may have changed in the meantime).
So if as a developer, you are not able to write a reliable test for this code, you have a third option: not testing it (acknowledging the fact that you have technical debt on this integration) instead of adding a test which test the wrong thing (and having the same technical debt but hiding it by fake tests).

and btw, Prophecy has always been described as an opinionated tool. You are still able to use another tool if the opinions block you there (using PHPunit mocks for this particular test for instance)

frankdejonge:
I'm not going to waste any more energy on this.

Taluu:
@stof IMO (and again, it's my opinion), I think the "don't mock what you don't own" shouldn't be applied as is. Especially in _unit_ tests ; the goal of these tests is to ensure that what _you_ made works, and fuck (stub / mock) everything else, 3rd party or not. Otherwise we would have to develop wrappers just to mock those ? It's not that I think this is a waste of time, but actually I think it is.

In integration / functionnal tests, there these 3rd party tools should indeed _not_ be mocked or whatever (this is where you test that you are correctly using the lib and other environmental parameters).

@frankdejonge too bad you closed the PR on the `_` :{

frankdejonge:
@Taluu the responses to the PR's I've made to this repository have only been replied to with theoretical arguments, without room for nuance, therefor I've decided not to contribute to this project anymore. I hadn't contributed yet anyway, so whatever. It was a small PR, people can just duplicate the effort.

PR #294, opened 2016-11-08
Remove (PHP 4 style) implicit private detection.

frankdejonge:
This is an alternative for #293. Instead of allowing yet another call, I've removed the older privacy convention detection.


tests/Doubler/Generator/ClassMirrorTest.php
```diff
@@ -9,20 +9,6 @@ class ClassMirrorTest extends \PHPUnit_Framework_TestCase
     /**
      * @test
      */
-    public function it_reflects_allowed_magic_methods()
```
stof: removing this test is wrong, as nothing ensures that they can be reflected anymore (and so forbidding them all would pass the tests)

tests/Doubler/Generator/ClassMirrorTest.php
```diff
@@ -301,23 +287,6 @@ public function it_reflects_all_interfaces_methods()
...
-        $this->assertCount(2, $classNode->getMethods());
-        $this->assertTrue($classNode->hasMethod('isAbstract'));
-        $this->assertTrue($classNode->hasMethod('__toString'));
-        $this->assertFalse($classNode->hasMethod('_getName'));
-    }
-
-    /**
-     * @test
-     */
     public function it_does_not_throw_exception_for_virtually_private_finals()
```
stof: does it still make sense ? You remove the concept of "virtually private"

frankdejonge:
@stof I've corrected it and removed the virtual private method fixtures.

PR #302, opened 2016-12-06
Allow \SoapClient's non-magic, magic-looking methods to be mocked.

bartfeenstra:
I'm trying to mock `\SoapClient::__soapCall()` in my own project, which looks like a magic method, but isn't. `\Prophecy\Doubler\Generator\ClassMirror::$reflectableMethods` lists the only methods starting with one (or more) underscore that Prophecy can mock. Adding `__soapCall` there makes my mock work as expected. I added the other non-magic, magic-looking sibling methods from http://php.net/manual/en/class.soapclient.php as well.

### DIFF
--- a/src/Prophecy/Doubler/Generator/ClassMirror.php
+++ b/src/Prophecy/Doubler/Generator/ClassMirror.php
@@ -32,7 +32,19 @@ class ClassMirror
         '__wakeup',
         '__toString',
         '__call',
-        '__invoke'
+        '__invoke',
+        // \SoapClient's non-magic methods that look like they are magic.
+        '__doRequest',
+        '__getFunctions',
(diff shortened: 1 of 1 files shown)


bartfeenstra:
Of course I did not find https://github.com/phpspec/prophecy/pull/293 or https://github.com/phpspec/prophecy/pull/294 until after I submitted my own PR.

jakzal:
Then you already read here https://github.com/phpspec/prophecy/pull/293#issuecomment-258853541 why this shouldn't be done ;)

everzet:
I am looking at possibility of removing class patches in the future altogether (maybe in next major). Mocking things you don't own, especially things that aren't really PHP, but C classes will always generate more pain than value.

PR #287, opened 2016-09-26
Add support for return type 'void'

prolic:
resolves: https://github.com/phpspec/prophecy/issues/280

PR #304, opened 2016-12-12
PHP 7.1 types

thomasvargiu:
Added support for PHP 7.1 types.

- Allowed `void` return type
- Allowed `iterable` type
- Allowed nullable arguments
- Allowed nullable return types

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ cache:
 branches:
   except:
     - /^bugfix\/.*$/
+    - /^feature\/.*$/
     - /^optimization\/.*$/
 
 matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ cache:
 branches:
   except:
     - /^bugfix\/.*$/
-    - /^feature\/.*$/
     - /^optimization\/.*$/
 
 matrix:
@@ -26,10 +25,10 @@ matrix:
       env: PHPDOCUMENTOR_REFLECTION_DOCBLOCK="^2.0"
 
 before_script:
+  - travis_retry composer update --no-interaction
   - if [ -n "$PHPDOCUMENTOR_REFLECTION_DOCBLOCK" ]; then
       composer require "phpdocumentor/reflection-docblock:${PHPDOCUMENTOR_REFLECTION_DOCBLOCK}" --no-update;
     fi;
-  - travis_retry composer update --no-interaction
 
 script:
   - vendor/bin/phpspec run -fpretty -v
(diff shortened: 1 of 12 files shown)


.travis.yml
```diff
@@ -26,10 +25,10 @@ matrix:
       env: PHPDOCUMENTOR_REFLECTION_DOCBLOCK="^2.0"
 
 before_script:
+  - travis_retry composer update --no-interaction
```
stof: this change is wrong. It must be done after changing dependencies (other the change is useless)

.travis.yml
```diff
@@ -11,7 +11,6 @@ cache:
 branches:
   except:
     - /^bugfix\/.*$/
-    - /^feature\/.*$/
```
stof: this must be reverted. We don't want such branches in the main repo to trigger Travis builds for the branch itself (they will be tested by the PR build)

composer.json
```diff
@@ -26,7 +26,7 @@
     },
 
     "require-dev": {
-        "phpspec/phpspec": "^2.0",
+        "phpspec/phpspec": "^2.0 || ^3.0",
```
stof: this is wrong, as the spec suite in the master branch is not compatible with phpspec 3 (there is a pending PR migrating to phpspec 3, but it clearly does not belong in this PR)

stof:
It looks like your PR contains the contributions of @prolic in #287 but under your own authorship. This is not good.

composer.json
thomasvargiu: Without this, tests doesn't work in PHP >= 5.6. Any solutions?

thomasvargiu:
@stof Oh, I didn't see it.  
@prolic, maybe I can integrate my other changes in you PR?

prolic:
Sure. Just submit a pr to my branch.

Am 13.12.2016 19:37 schrieb "Thomas Mauro Vargiu" <notifications@github.com

thomasvargiu:
@stof @prolic. I think my PR is more complete with PHP 7.1 support.
#287 allows only void return type, this PR allows all PHP 7.1 changes: `void` return type, nullable arguments, nullable return types, `iterable` type.

@prolic what do you think?

thomasvargiu:
Ok, I created PR prolic/prophecy#1.

Closed in favor of #287

composer.json
thomasvargiu: @ciaranmcnulty We have the same issue in #287 . You can see on this build: https://travis-ci.org/phpspec/prophecy/builds/183575285

ciaranmcnulty: While it would make sense to upgrade the suite to work with PhpSpec 3, this is a bug in that test, in prophecy, or an edge case in phpspec that should be fixed. PhpSpec 2 is still in bugfix support for another 6 months

ciaranmcnulty: this is broken on master - I'll take a look

Issue #305, opened 2016-12-13
A dependency has caused the test suite to break


ciaranmcnulty:
The build on master 22 days ago passed, today it's breaking.

This is reproducible locally, but doing `composer update --prefer-lowest` removes the error. This suggests the break is down to one of our dependencies introducing a change that broke prophecy.

More work is needed to figure out which one...

thomasvargiu:
@ciaranmcnulty @sebastianbergmann 

The problem was introduced with PR #298.

Doing a `composer update`, `phpspec/phpspec` 2.0.1 is installed, because doesn't require `sebastian/exporter`.

To use `sebastian/recursion-context` ^2.0 I think we should use `phpspec/phpspec` ^3.2.2.

ciaranmcnulty:
Why did travis not catch it?

thomasvargiu:
Specs are not compatible with phpspec 2.0 I think. We should make tests with `--prefer-lowest` to ensure minimum compatibility.

PR #306, opened 2016-12-13
Bumped phpspec to latest version

thomasvargiu:
Fix for #305

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,13 +22,6 @@ matrix:
       sudo: required
       dist: trusty
       group: edge
-    - php: '7.0'
-      env: PHPDOCUMENTOR_REFLECTION_DOCBLOCK="^2.0"
-
-before_script:
-  - if [ -n "$PHPDOCUMENTOR_REFLECTION_DOCBLOCK" ]; then
-      composer require "phpdocumentor/reflection-docblock:${PHPDOCUMENTOR_REFLECTION_DOCBLOCK}" --no-update;
-    fi;
 
 install:
   - travis_retry composer update --no-interaction
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,10 +14,6 @@ branches:
     - /^feature\/.*$/
     - /^optimization\/.*$/
 
-env:
-  global:
-    - COMPOSER_ARGS="--prefer-stable --no-interaction"
-
 matrix:
   fast_finish: true
   include:
@@ -30,7 +26,7 @@ matrix:
       env: PHPDOCUMENTOR_REFLECTION_DOCBLOCK="^2.0"
     - php: 7.1
       env:
-        - COMPOSER_ARGS="--prefer-stable --prefer-lowest --no-interaction"
+        - COMPOSER_ARGS="--prefer-lowest --no-interaction"
 
 before_install:
   - phpenv config-rm xdebug.ini || return 0
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,10 @@ before_script:
   - if [ -n "$PHPDOCUMENTOR_REFLECTION_DOCBLOCK" ]; then
       composer require "phpdocumentor/reflection-docblock:${PHPDOCUMENTOR_REFLECTION_DOCBLOCK}" --no-update;
     fi;
+
+install:
   - travis_retry composer update --no-interaction
+  - composer show --installed
 
 script:
   - vendor/bin/phpspec run -fpretty -v
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,51 +14,19 @@ branches:
 
 env:
   global:
-    - COMPOSER_ARGS="--prefer-stable --no-interaction"
+    - COMPOSER_ARGS="--no-interaction"
+    - DEPS=latest
 
 matrix:
   fast_finish: true
   include:
     - php: 5.3
-      env:
-        - DEPS=lowest
-    - php: 5.3
-      env:
-        - DEPS=latest
     - php: 5.4
-      env:
-        - DEPS=lowest
-    - php: 5.4
-      env:
-        - DEPS=latest
     - php: 5.5
-      env:
-        - DEPS=lowest
-    - php: 5.5
-      env:
-        - DEPS=latest
     - php: 5.6
-      env:
-        - DEPS=lowest
-    - php: 5.6
-      env:
-        - DEPS=latest
     - php: 7.0
-      env:
-        - DEPS=lowest
-    - php: 7.0
-      env:
-        - DEPS=latest
     - php: 7.1
-      env:
-        - DEPS=lowest
     - php: 7.1
-      env:
-        - DEPS=latest
-    - php: hhvm
-      sudo: required
-      dist: trusty
-      group: edge
       env:
         - DEPS=lowest
     - php: hhvm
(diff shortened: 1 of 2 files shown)


thomasvargiu:
Only hhvm enviroment failed. I think it should pass on master branch.

.travis.yml
```diff
@@ -22,14 +22,10 @@ matrix:
       sudo: required
       dist: trusty
       group: edge
-    - php: '7.0'
-      env: PHPDOCUMENTOR_REFLECTION_DOCBLOCK="^2.0"
 
-before_script:
-  - if [ -n "$PHPDOCUMENTOR_REFLECTION_DOCBLOCK" ]; then
-      composer require "phpdocumentor/reflection-docblock:${PHPDOCUMENTOR_REFLECTION_DOCBLOCK}" --no-update;
```
stof: Removing testing against multiple versions of ``phpdocumentor/reflection-docblock`` is a no-go IMO. We want to test our support of phpdocumentor/reflection-docblock 2, as we want to support both 2.x and 3.x in parallel.

.travis.yml
```diff
@@ -22,14 +22,10 @@ matrix:
...
-    fi;
+install:
   - travis_retry composer update --no-interaction
+  - composer show --installed
```
stof: this is useless. The output of ``composer update`` already allows to show what gets installed.

.travis.yml
```diff
@@ -14,18 +12,66 @@ branches:
...
+        - DEPS=lowest
+    - php: 5.4
+      env:
+        - DEPS=latest
+    - php: 5.5
+      env:
+        - DEPS=lowest
+    - php: 5.5
+      env:
+        - DEPS=latest
```
stof: Please revert this. Testing lowest deps on all PHP versions makes the matrix huge, and we have a quota of concurrent jobs per github organization on Travis, so huge matrix have drawbacks.

.travis.yml
```diff
@@ -14,18 +12,66 @@ branches:
...
       group: edge
+      env:
+        - DEPS=latest
 
 install:
-  - travis_retry composer update --no-interaction
-  - composer show --installed
+  - if [[ $DEPS == 'latest' ]]; then travis_retry composer update $COMPOSER_ARGS ; fi
+  - if [[ $DEPS == 'lowest' ]]; then travis_retry composer update --prefer-lowest $COMPOSER_ARGS ; fi
+  - travis_retry composer install $COMPOSER_ARGS
```
stof: running install after the update is useless

.travis.yml
thomasvargiu: @stof What would you suggest? Should we test lowest dependecies only in one PHP version?

.travis.yml
```diff
@@ -1,7 +1,5 @@
 language: php
 
-php: [5.3, 5.4, 5.5, 5.6, 7.0, 7.1]
```
stof: I would revert this. HAving the list here is more readable than including them explicitly each time

.travis.yml
```diff
@@ -14,22 +12,37 @@ branches:
...
     - php: hhvm
       sudo: required
       dist: trusty
       group: edge
-    - php: '7.0'
-      env: PHPDOCUMENTOR_REFLECTION_DOCBLOCK="^2.0"
-
-before_script:
-  - if [ -n "$PHPDOCUMENTOR_REFLECTION_DOCBLOCK" ]; then
-      composer require "phpdocumentor/reflection-docblock:${PHPDOCUMENTOR_REFLECTION_DOCBLOCK}" --no-update;
```
stof: Once again, the testing against `` PHPDOCUMENTOR_REFLECTION_DOCBLOCK="^2.0"`` MUST be kept.

.travis.yml
```diff
@@ -14,22 +12,37 @@ branches:
...
-    fi;
-  - travis_retry composer update --no-interaction
+      env:
+        - DEPS=latest
+
+before_install:
+  - phpenv config-rm xdebug.ini || return 0
+  - composer self-update
```
stof: no need to self-update. Travis does it already

.travis.yml
```diff
@@ -14,22 +12,37 @@ branches:
...
 
+env:
+  global:
+    - COMPOSER_ARGS="--no-interaction"
+    - DEPS=latest
+
 matrix:
   fast_finish: true
   include:
-    # Use the newer stack for HHVM as HHVM does not support Precise anymore since a long time and so Precise has an outdated version
```
stof: please keep this comment above the code including HHVM.

.travis.yml
```diff
@@ -14,22 +12,37 @@ branches:
...
+    - php: 5.3
+    - php: 5.4
+    - php: 5.5
+    - php: 5.6
+    - php: 7.0
+    - php: 7.1
+    - php: 7.1
+      env:
+        - DEPS=lowest
```
stof: rather than ``deps=lowest``, I would overwrite the ``COMPOSER_FLAGS`` variable here, avoiding the need to have different commands for the ``composer update`` call

.travis.yml
thomasvargiu: Why this? lowest deps already use the minimum version of dockblock (2.0).

.travis.yml
thomasvargiu: Without the list, we can't test with latest and lowest dependencies on the same PHP version. We should keep the list IMO.

stof: you can keep this list and include additional versions. This is what we were doing already for dockblock 2.x

thomasvargiu: IMO, I think that now that we are testing with lowest dependencies, it is not necessary to test the last version of 2.x.

Usually a project is tested with lowest and latest dependencies version and assumes that dependencies are following [semver](http://semver.org). So no BC break should be exist between minor changes.

This is why I think that is enough to test with lowest dependencies.

What do you think?

stof: Well, the latest 2.x version is much more likely to be used by our users than the lowest bound.

And this does not change the fact that I prefer keeping the list of php versions here rather than listing them all in ``include`` for default configurations

thomasvargiu: Ok. Can you check last commit?

Other builds are:
- PHP 7.0 and docblock ^2.0
- PHP 7.1 and `--prefer-lowest`

If you want, I can remove the `--prefer-lowest`, but I think would be useful.

.travis.yml
```diff
@@ -14,6 +14,10 @@ branches:
     - /^feature\/.*$/
     - /^optimization\/.*$/
 
+env:
+  global:
+    - COMPOSER_ARGS="--prefer-stable --no-interaction"
```
stof: ``--no-interaction`` can be removed entirely, as Travis forces it through the env variable anyway.

And ``--prefer-stable`` is useless here, as the min stability is stable anyway (and it changes the existing behavior)

thomasvargiu: I've removed it. Thanks.

thomasvargiu:
@stof If it's ok for you, is it possible to merge it?

9orky:
This is just a memo for @stof to take a look at this PR and merge it :)

ruudk:
Any news on this 🙏 ? Would love to use Prophecy on PHP 7.1 (with #287)

sandrokeil:
@stof Please merge :-)

MidnightDesign:
Is there any help needed? Would be awesome if we could merge this as soon as possible.

everzet:
Done and merged through another PR. Thanks for the hard work, though!

PR #308, opened 2016-12-19
Discover api from PHPDoc of *all* parent classes

webimpress:
Example:

```
/**
 * @method void undefinedMethod()
 */
class MagicalApi
{
}

/**
 * @method void undefinedMethodChild()
 */
class MagicalApiChild extends MagicalApi
{
}
```

Should be possible mock methods defined in child and parent PHPDocs:
- `undefinedMethod`
- `undefinedMethodChild`

So:
```php
$prophesize = $this->prophesize(MagicalApiChild::class);
$prophesize->undefinedMethod()->willReturn(...);
$prophesize->undefinedMethodChild()->willReturn(...);
```

Currently is possible to mock only `undefinedMethodChild`:
```php
$prophesize = $this->prophesize(MagicalApiChild::class);
$prophesize->undefinedMethodChild()->willReturn(...);
```
and:

```php
$prophesize->undefinedMethod()->willReturn(...);
```

will throw an exception:
```
Method `Double\MagicalApiChild\P2::undefinedMethod()` is not defined.
```

### DIFF
--- a/spec/Prophecy/Doubler/ClassPatch/MagicCallPatchSpec.php
+++ b/spec/Prophecy/Doubler/ClassPatch/MagicCallPatchSpec.php
@@ -29,6 +29,17 @@ class MagicCallPatchSpec extends ObjectBehavior
         $this->apply($node);
     }
 
+    function it_discovers_api_using_phpdocs_of_parent_classes(ClassNode $node)
+    {
+        $node->getParentClass()->willReturn('spec\Prophecy\Doubler\ClassPatch\MagicalApiChild');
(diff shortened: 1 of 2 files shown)


everzet:
Sorry for a very late reply. I don't really want to add more magic into the core. If anything, we already cover too many challenges created due to a single breakage of a cardinal rule - don't mock things you don't own.

Virtual interfaces and classes are better mocked/faked by other means. And if you don't own the virtual interface in the first place, change of approach might be a better match.

PR #309, opened 2017-01-02
Allowing reserved keywords, since PHP 7.0

MartinPham:
Starting from PHP 7.0, some keywords can be used. So we exclude from list of keywords, and only add when PHP version < 7.0

### DIFF
--- a/src/Prophecy/Doubler/ClassPatch/KeywordPatch.php
+++ b/src/Prophecy/Doubler/ClassPatch/KeywordPatch.php
@@ -61,8 +61,7 @@ class KeywordPatch implements ClassPatchInterface
      * @return array
      */
     private function getKeywords() {
-
-        return array(
+        $keywords = array(
             '__halt_compiler',
             'abstract',
             'and',
@@ -108,7 +107,6 @@ class KeywordPatch implements ClassPatchInterface
             'insteadof',
             'interface',
             'isset',
-            'list',
             'namespace',
             'new',
             'or',
@@ -131,5 +129,19 @@ class KeywordPatch implements ClassPatchInterface
             'xor',
             'yield',
         );
+
+        /*
+         * Starting from PHP 7.0, some keywords can be used
+         * So we exclude from list of keywords, and only add
+         * when PHP version < 7.0
+         *
+         * https://wiki.php.net/rfc/context_sensitive_lexer
+         */
+        if(version_compare(PHP_VERSION, '7.0', '<'))
+        {
+            $keywords[] = 'list';
+        }
+
+        return $keywords;
     }
 }
--- a/src/Prophecy/Doubler/ClassPatch/KeywordPatch.php
+++ b/src/Prophecy/Doubler/ClassPatch/KeywordPatch.php
@@ -61,87 +61,84 @@ class KeywordPatch implements ClassPatchInterface
      * @return array
      */
     private function getKeywords() {
-        $keywords = array(
-            '__halt_compiler'
-        );
-
         /*
          * Starting from PHP 7.0, almost keywords can be used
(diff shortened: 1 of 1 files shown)


stof:
I think your list of keywords being allowed on PHP 7 is incomplete here. Many more keywords are now allowed in method names AFAIK.

I'm in favor of accepting such changes, but only if done in a complete way, not only for ``list``

MartinPham:
@stof I've updated the code, thanks!

src/Prophecy/Doubler/ClassPatch/KeywordPatch.php
```diff
@@ -61,75 +61,87 @@ public function getPriority() {
...
-            'trait',
-            'try',
-            'unset',
-            'use',
-            'var',
-            'while',
-            'xor',
-            'yield',
+        $keywords = array(
+            '__halt_compiler'
```
stof: is this really the only forbidden keyword on PHP 7 ? If yes, I would rather do this:

```php
if (\PHP_VERSION_ID >= 70000) {
    return array('__halt_compiler');
}

return array(
    '__halt_compiler',
    // ...
);
```

It would keep the code much more readable.

MartinPham: Yea, I thought about it also (it is bad to append keywords many times), but I'm worry next versions will have some more reserved keywords.

MartinPham:
@stof I've modified it based on your idea, thanks!

src/Prophecy/Doubler/ClassPatch/KeywordPatch.php
```diff
@@ -130,6 +139,6 @@ private function getKeywords() {
             'while',
             'xor',
             'yield',
-        );
+	    );
```
stof: can you revert this change ?

src/Prophecy/Doubler/ClassPatch/KeywordPatch.php
```diff
@@ -61,6 +61,15 @@ public function getPriority() {
      * @return array
      */
     private function getKeywords() {
+        /*
+         * Starting from PHP 7.0, almost keywords can be used
```
avant1: Probably typo: "almost **all** keywords"

MartinPham:
@stof @avant1 done :)

src/Prophecy/Doubler/ClassPatch/KeywordPatch.php
```diff
@@ -129,7 +138,7 @@ private function getKeywords() {
             'var',
             'while',
             'xor',
-            'yield',
-        );
+            'yield'
+	);
```
avant1: I think stof wanted this line to be not changed in this PR.
Now there is tab character instead of 4 spaces.

avant1:
@MartinPham can you rebase this changes on master? Build should be fixed in master now.

everzet:
Please rebase and try to fix the build.

src/Prophecy/Doubler/ClassPatch/KeywordPatch.php
```diff
@@ -129,7 +138,7 @@ private function getKeywords() {
             'var',
             'while',
             'xor',
-            'yield',
-        );
+            'yield'
```
stof: the trailing comma should not be removed either

nesl247:
Any chance we can get this merged in? Just ran into this issue and it's a big blocker for me.

olvlvl:
WANT! I just created this issue: https://github.com/phpspec/prophecy/issues/351

Taluu:
was already taken care of in https://github.com/phpspec/prophecy/pull/337

Issue #303, opened 2016-12-08
Nullable return method "must be an instance of {class name}, null returned"


elegos:
PHP 7.1.0 (release)
PHPUnit 5.7.2
Prophecy 1.6.2

Hello!

With the new PHP 7.1.0 released some days ago, PHP introduced the nullable return. I've tried mocking one of these methods, but apparently Prophecy does not support them yet (PHPUnit with `PHPUnit_Framework_MockObject_MockObject` does, so it's not PHPUnit).

May you please take a look at it?

This is a minimal test which reproduces the error:

    <?php
    
    class Nullable {
        public function getNullable() : ?Nullable {
            return null;
        }
    }
    
    class BaseProphecyFailTest extends PHPUnit_Framework_TestCase
    {
        public function testNullable()
        {
            /** @var Nullable|\Prophecy\Prophecy\ObjectProphecy $mock */
            $mock = $this->prophesize(Nullable::class);
            $mock->getNullable()->willReturn(null);
    
            $mock = $mock->reveal();
    
            $this->assertNull($mock->getNullable());
        }
    }

Output:

    Testing started at 09:50 ...
    PHPUnit 5.7.2 by Sebastian Bergmann and contributors.

    Return value of Double\Nullable\P1::getNullable() must be an instance of Nullable, null returned
     BaseProphecyFailTest.php:19

    Time: 121 ms, Memory: 4.00MB

    ERRORS!
    Tests: 1, Assertions: 0, Errors: 1.
    
    Process finished with exit code 2

deeky666:
I think this is fixed by #322 and can be closed.

PR #311, opened 2017-01-18
Fix issue #303

jojo1981:
Fixed: Nullable return method "must be an instance of {class name}, null returned"

### DIFF
--- a/tests/Doubler/Generator/ClassMirrorTest.php
+++ b/tests/Doubler/Generator/ClassMirrorTest.php
@@ -331,14 +331,12 @@ class ClassMirrorTest extends \PHPUnit_Framework_TestCase
 
     /**
      * @test
-     * @requires PHP 7
+     * @requires PHP 7.1
      */
-    public function it_reflects_return_typehints()
+    public function it_reflects_return_nullable_typehints()
     {
-        $class = new \ReflectionClass('Fixtures\Prophecy\WithReturnTypehints');
-
+        $class = new \ReflectionClass('Fixtures\Prophecy\WithNullableReturnTypes');
         $mirror = new ClassMirror();
-
         $classNode = $mirror->reflect($class, array());
 
         $this->assertCount(3, $classNode->getMethods());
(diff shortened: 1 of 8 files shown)


fixtures/WithReturnTypehints70.php
```diff
@@ -2,7 +2,7 @@
 
 namespace Fixtures\Prophecy;
 
-class WithReturnTypehints extends EmptyClass
+class WithReturnTypehints70 extends EmptyClass
```
stof: why adding 70 here ? This class is not specific to PHP 7.0

fixtures/WithReturnTypehints70.php
```diff
@@ -15,4 +15,8 @@ public function getName(): string {
     public function getParent(): parent {
         return $this;
     }
+
+    public function getNullableReturnType() : \stdClass {
```
stof: this is *not* a nullable return type, so this is a lie

fixtures/WithReturnTypehints71.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace Fixtures\Prophecy;
+
+class WithReturnTypehints71 extends EmptyClass
```
stof: You should instead have a ``WithNullableReturnTypes`` class containing only methods with nullable types, and used by **different** tests that the existing class (and these tests being skipped before PHP 7.1)

tests/Doubler/Generator/ClassMirrorTest.php
```diff
@@ -335,20 +335,48 @@ public function it_does_not_throw_exception_for_virtually_private_finals()
      */
     public function it_reflects_return_typehints()
     {
-        $class = new \ReflectionClass('Fixtures\Prophecy\WithReturnTypehints');
+        if (version_compare(PHP_VERSION, '7.1', '>=')) {
+            $class = new \ReflectionClass('Fixtures\Prophecy\WithReturnTypehints71');
+        } else {
+            $class = new \ReflectionClass('Fixtures\Prophecy\WithReturnTypehints70');
```
stof: this is a no-go

tests/Doubler/Generator/ClassMirrorTest.php
```diff
@@ -335,20 +335,48 @@ public function it_does_not_throw_exception_for_virtually_private_finals()
...
+        $this->assertFalse($classNode->getMethod('getSelf')->isReturnTypeNullable());
+
         $this->assertEquals('\Fixtures\Prophecy\EmptyClass', $classNode->getMethod('getParent')->getReturnType());
+        $this->assertFalse($classNode->getMethod('getParent')->isReturnTypeNullable());
+
+        $this->assertEquals('\stdClass', $classNode->getMethod('getNullableReturnType')->getReturnType());
+        if (version_compare(PHP_VERSION, '7.1', '>=')) {
+            $this->assertTrue($classNode->getMethod('getNullableReturnType')->isReturnTypeNullable());
+        } else {
+            $this->assertFalse($classNode->getMethod('getNullableReturnType')->isReturnTypeNullable());
```
stof: this does not make any sense

fixtures/WithReturnTypehints70.php
jojo1981: I agree this is not OK I will fix it.

fixtures/WithReturnTypehints70.php
jojo1981: Agree, also this I will fix.

fixtures/WithReturnTypehints71.php
jojo1981: I agree this is a far better solution. So I will fix it.

tests/Doubler/Generator/ClassMirrorTest.php
jojo1981: Agree I will fix it and have only a PHP 7.1 check.

tests/Doubler/Generator/ClassMirrorTest.php
jojo1981: I will fix it.

unfunco: `PHP_VERSION_ID` is available as of 5.2.7 too, and Prophecy supports 5.3+ so it's safe to use, and will have slightly better performance when compared with `version_compare`.

fixtures/WithReturnTypehints71.php
jojo1981: Agree. Added a new test and make use of the PHP Unit `@requires` annotation.

tests/Doubler/Generator/ClassMirrorTest.php
jojo1981: I have reverted this and added a new nullable return types test.

jojo1981:
@stof thanks for the review. I have processed all the comments and updated this PR. Before pushing I had squashed the commits.

jojo1981:
@stof can you please check this PR again?

fgheorghe:
Would be great to have this reviewed as it is blocking proper testing of nullable methods.

Sam-Burns:
Other than the need fix the `composer.json` conflict, it seems like this PR is mergeable now.  It looks like all the feedback from the code review has been addressed, and it would be good to have working support for nullable types.

everzet:
Please rebase.

jojo1981:
This PR has no purpose anymore because this PR is merged: https://github.com/phpspec/prophecy/pull/322

PR #313, opened 2017-02-01
Break external references to Call arguments

### LINKED ISSUE #75: Problem with predictions check
As it stands now, predictions are checked on teardown. Since call center records calls with reference to arguments, operations on these arguments after the call lead to unexpected results.
Here's an example that gives me this issue:

```
// example
function it_makes_proper_collaborator_call($collaborator)
{
   $collaborator->execute(Argument::which('getName', 'A'))->shouldBeCalled();
   $this->run();
}

//implementation
function run()
{
  $event = new Event();
  $event->setName('A');
  $this->collaborator->execute($event);
  $event->setName('B');
}
```

It fails, because when predictions are checked the `$event` has wrong `name`.

danizord:
Attempt to fix #75

### DIFF
--- a/spec/Prophecy/Call/CallSpec.php
+++ b/spec/Prophecy/Call/CallSpec.php
@@ -3,6 +3,7 @@
 namespace spec\Prophecy\Call;
 
 use PhpSpec\ObjectBehavior;
+use stdClass;
 
 class CallSpec extends ObjectBehavior
 {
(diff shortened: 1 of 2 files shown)


stof:
This has 2 issues:

- passing non-serializable objects is a necessary use case
- this breaks the ``ExactToken`` matching when using a spy

danizord:
Hey @stof! Is there any other way to solve this issue?

stof:
there is no *easy* way to solve it

everzet:
This one is part of some of my nightmares, sadly. My guess would be that the solution would require major rethink of couple of core components, and, hence the major version bump.

PR #317, opened 2017-02-23
Allow sebastian/recursion-context ^3.0

sebastianbergmann:
sebastian/recursion-context ^3.0 will not break API. It just drops support for PHP 5.


sebastianbergmann:
Thank you so much!

PR #318, opened 2017-02-24
Fixing build

Sam-Burns:


.travis.yml
```diff
@@ -15,21 +13,25 @@ branches:
     - /^optimization\/.*$/
 
 matrix:
-  fast_finish: true
   include:
-    # Use the newer stack for HHVM as HHVM does not support Precise anymore since a long time and so Precise has an outdated version
```
stof: please keep this comment above the HHVM special config

.travis.yml
```diff
@@ -15,21 +13,25 @@ branches:
...
+      env: DEPENDENCIES='low'
+    - php: 5.6
+    - php: 7.0
+    - php: 7.1
     - php: hhvm
       sudo: required
       dist: trusty
       group: edge
-    - php: '7.0'
-      env: PHPDOCUMENTOR_REFLECTION_DOCBLOCK="^2.0"
```
stof: please don't remove this job. We want to test against the latest version of reflection-docblock 2.x

.travis.yml
```diff
@@ -15,21 +13,25 @@ branches:
...
 
-before_script:
-  - if [ -n "$PHPDOCUMENTOR_REFLECTION_DOCBLOCK" ]; then
-      composer require "phpdocumentor/reflection-docblock:${PHPDOCUMENTOR_REFLECTION_DOCBLOCK}" --no-update;
-    fi;
-  - travis_retry composer update --no-interaction
+install:
+  - export COMPOSER_ROOT_VERSION=dev-master
+  - if [ "$DEPENDENCIES" != "low" ]; then composer update; fi;
+  - if [ "$DEPENDENCIES" == "low" ]; then composer update --prefer-lowest; fi;
```
stof: to simplify this, I generally do ``COMPOSER_FLAGS='--prefer-lowest'`` instead of ``DEPENDENCIES='low'``, and the ``composer update $COMPOSER_FLAGS``

.travis.yml
```diff
@@ -1,7 +1,5 @@
 language: php
 
-php: [5.3, 5.4, 5.5, 5.6, 7.0, 7.1]
```
stof: I suggest keeping this list for all jobs not needing a special config, instead of adding them in ``matrix > include``

composer.json
```diff
@@ -26,8 +26,8 @@
     },
 
     "require-dev": {
-        "phpspec/phpspec": "^2.0",
-        "phpunit/phpunit": "^4.8 || ^5.6.5"
+        "phpspec/phpspec": "^2.4.0",
+        "phpunit/phpunit": "^4.8 || ^5.6.5 || ^6.0"
```
stof: that's wrong. Our testsuite is not yet compatible with PHPUnit 6, so this is *breaking* the build

Sam-Burns: @stof the build isn't broken.

.travis.yml
Sam-Burns: It's copied straight out of `phpspec/phpspec`.  Why not just keep them consistent?

composer.json
stof: That's because for some reason, Composer always install 4.8.35 even on PHP 7.1

.travis.yml
ciaranmcnulty: Different projects :-)

composer.json
stof: I understood why. PhpSpec 2 is not compatible with ``sebastian/exporter`` 2, meaning it is incompatible with PHPUnit 5+

.travis.yml
stof: my own advice would then be to simplify the travis.yml config in phpspec (which itself copied it from Symfony, where there is a valid reason to use ``$DEPENDENCIES`` as it changes a bunch of other commands depending on builds)

Sam-Burns: No, I don't think I'll change that.  If it's good enough for PhpSpec, Symfony, PHPUnit and Behat, I'm sure this approach will work fine for Prophecy.

.travis.yml
Sam-Burns: It wasn't testing anything... If you look at the [composer command](https://github.com/phpspec/prophecy/blob/e1f39368ce912e031df11ebecfc23be12007701d/.travis.yml#L30) being used, it ran with the `--no-update` flag.  This means it didn't change anything.

sstok: `--no-update` means Composer doesn't run update but only updates the file contents.

PR #322, opened 2017-03-02
Add PHP 7.1 Support

Sam-Burns:
Revival of @prolic's excellent work on https://github.com/phpspec/prophecy/pull/287.  Changes:
- Rebased
- Adds commit fixing build failure caused by use of void return types in class generation with PHP 7.0.

### DIFF
--- a/spec/Prophecy/Doubler/Generator/ClassCodeGeneratorSpec.php
+++ b/spec/Prophecy/Doubler/Generator/ClassCodeGeneratorSpec.php
@@ -16,6 +16,7 @@ class ClassCodeGeneratorSpec extends ObjectBehavior
         MethodNode $method1,
         MethodNode $method2,
         MethodNode $method3,
+        MethodNode $method4,
         ArgumentNode $argument11,
         ArgumentNode $argument12,
         ArgumentNode $argument21,
@@ -26,7 +27,7 @@ class ClassCodeGeneratorSpec extends ObjectBehavior
             'Prophecy\Doubler\Generator\MirroredInterface', 'ArrayAccess', 'ArrayIterator'
         ));
         $class->getProperties()->willReturn(array('name' => 'public', 'email' => 'private'));
-        $class->getMethods()->willReturn(array($method1, $method2, $method3));
+        $class->getMethods()->willReturn(array($method1, $method2, $method3, $method4));
 
         $method1->getName()->willReturn('getName');
         $method1->getVisibility()->willReturn('public');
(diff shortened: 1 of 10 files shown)


spec/Prophecy/Doubler/Generator/ClassCodeGeneratorSpec.php
```diff
@@ -59,30 +63,54 @@ function it_generates_proper_php_code_for_specific_ClassNode(
...
+private $email;
+
+public static function getName(array $fullname = NULL, \ReflectionClass $class): ?string {
+return $this->name;
+}
+protected  function getEmail(?string $default = 'ever.zet@gmail.com') {
+return $this->email;
+}
+public  function &getRefValue( $refValue): void {
+return $this->refValue;
```
stof: this is invalid code

src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
```diff
@@ -60,27 +60,60 @@ private function generateMethod(Node\MethodNode $method)
...
+        if (version_compare(PHP_VERSION, '7.1', '>=')) {
+            if ($method->hasReturnType()) {
+                return $method->hasNullableReturnType()
+                    ? sprintf(': ?%s', $method->getReturnType())
+                    : sprintf(': %s', $method->getReturnType());
+            }
+        }
+
+        if (version_compare(PHP_VERSION, '7.0', '>=')) {
+            return $method->hasReturnType() && $method->getReturnType() !== 'void'
```
stof: why doing a special case for void here ?

src/Prophecy/Doubler/Generator/ClassMirror.php
```diff
@@ -155,6 +155,10 @@ private function reflectMethodToNode(ReflectionMethod $method, Node\ClassNode $c
             }
 
             $node->setReturnType($returnType);
+
+            if (version_compare(PHP_VERSION, '7.1', '>=') && $method->getReturnType()->allowsNull()) {
+                $node->setNullableReturnType(true);
```
stof: please add a new integration test in the PHPUnit testsuite covering this case

src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
Sam-Burns: There was an error where void return types were being added in PHP 7.0, which was causing a build break.

src/Prophecy/Doubler/Generator/ClassMirror.php
Sam-Burns: The `ClassCodeGeneratorSpec` seems to have this covered already

spec/Prophecy/Doubler/Generator/ClassCodeGeneratorSpec.php
Sam-Burns: The return statement?  Why?

danizord: @Sam-Burns Function signature specifies void return type.

Sam-Burns: Ah, I see what you mean

everzet:
Thanks for a great work @Sam-Burns and especially @prolic. Sorry for a long wait

sukei:
👍 Thanks you guys !

src/Prophecy/Doubler/Generator/ClassMirror.php
stof: No. The spec for ClassCodeGeneratorSpec does not test the ClassMirror at all.

src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
```diff
@@ -60,27 +60,60 @@ private function generateMethod(Node\MethodNode $method)
...
         return $php.'}';
     }
 
+    /**
+     * @return string
+     */
+    private function getReturnType(Node\MethodNode $method)
+    {
+        if (version_compare(PHP_VERSION, '7.1', '>=')) {
+            if ($method->hasReturnType()) {
```
Taluu: why was this condition split ?

src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
Taluu: imo if the node has a `void` return type and it is on php 7.0. it should fail anyway (it's up to the tests to skip said test if not php 7.1+)...

Here, somebody having a `void` return type on a node and php 7.0 will not have the method properly defined, and so could lead to errors.

PR #325, opened 2017-04-05
Fixed typo in MethodName case usage

mentor:
Fellow Prophecy devs,

I’ve fixed Bug in MethodName.php => on line 119 **case 'string';** semicolon was used instead of **case 'string':** colon usage. I have fixed it in the branch named “semicolon-fix” in my fork. Thanks, and I hope you will find my pull-request reasonable enough to be merged to master.

Cheers,
Mentor

### DIFF
--- a/src/Prophecy/Doubler/Generator/Node/MethodNode.php
+++ b/src/Prophecy/Doubler/Generator/Node/MethodNode.php
@@ -117,7 +117,7 @@ class MethodNode
                 $this->returnType = null;
                 break;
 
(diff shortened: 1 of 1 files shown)


everzet:
It would be great if you can provide a failing test case, to prove your fix works :)

stof:
This is just a coding standard issue, not a bug. PHP allows both semi-colon and colon after the case statement: http://php.net/manual/en/control-structures.switch.php

PR #327, opened 2017-04-07
feature: DateTimeToken for datetime with delta related assertions

Markcial:
A DateTime Token that allows to check if a argument received on a function is a datetime inside a delta precision measured on seconds.

### DIFF
--- a/src/Prophecy/Argument/Token/DateTimeDeltaToken.php
+++ b/src/Prophecy/Argument/Token/DateTimeDeltaToken.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace Prophecy\Argument\Token;
+
+
+class DateTimeDeltaToken implements TokenInterface
+{
+    /** @var \DateTime */
+    private $value;
+    /** @var integer */
+    private $delta;
+
+    public function __construct(\DateTimeInterface $value, $delta)
+    {
+        $this->value = $value;
+        $this->delta = $delta;
+    }
+
+    /**
+     * Calculates token match score for provided argument.
+     *
+     * @param $argument
+     *
+     * @return bool|int
+     */
+    public function scoreArgument($argument)
+    {
+        if (!$argument instanceof \DateTimeInterface) {
+            return false;
+        }
+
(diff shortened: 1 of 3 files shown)


everzet:
`scoreArgument` should return an `int` on positive match. The higher the `int`, the higher the match ranked during comparison.

See more here: https://github.com/phpspec/prophecy/blob/master/src/Prophecy/Argument/ArgumentsWildcard.php#L43-L76

src/Prophecy/Argument/Token/DateTimeDeltaToken.php
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * Calculates token match score for provided argument.
+     *
+     * @param $argument
+     *
+     * @return bool|int
+     */
+    public function scoreArgument($argument)
+    {
+        if (!$argument instanceof \DateTime) {
```
stof: this won't allow working with DateTimeImmutable

Markcial: I'll add support for the inmutable type too.

Markcial:
@everzet the int value should be based on the difference on milliseconds between the two arguments? for the 7.1 < versions should be ok, but if i did this branch was because we have som tests failing because of the newer microseconds support on datetimes for PHP7.1. (see : http://php.net/manual/en/migration71.incompatible.php)

How i should address this? set the difference as microseconds but this difference only will be noticeable on php7.1 >= versions, for the older versions it will display multiples of a thousand if there is a difference supperior of a millisecond.

everzet:
You misunderstood me. Every Token returns [integer constant on match](https://github.com/phpspec/prophecy/blob/master/src/Prophecy/Argument/Token/AnyValueToken.php#L30). That integer represents priority of the token and is used by Wildcard to [score and compare](https://github.com/phpspec/prophecy/blob/master/src/Prophecy/Argument/ArgumentsWildcard.php#L64) argument sets.

Higher that integer value is, the more likely argument set with your token will win. Rule of thumb at the moment is that more precise tokens score higher integers (to ensure they win).

PR #328, opened 2017-04-13
Add gitattributes file

jdufresne:
Used by composer to ignore tests and other development files in the
distribution of a release.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,8 @@
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.travis.yml export-ignore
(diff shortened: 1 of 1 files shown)


sstok:
spec directory should also be ignored. But not sure if the developers want to keep this as examples 😃

jdufresne:
Thanks for the feedback. Initially, I wasn't sure if `spec` was used at runtime. It has been ignored in the version. Just my opinion, but I think the `vendor` directory isn't a good location for docs and examples. They're better suited as online material.

everzet:
@jdufresne thanks for contribution!

PR #331, opened 2017-05-02
@see doctag must be fqn for ide navigation

Haehnchen:
To make `@see` navigable it must be fqn with a starting backslash


everzet:

PR #336, opened 2017-05-25
README.md: update code snippet for a consistency

petrkotek:
Update `->willThrow` to `->willThrow($exception)`, because other `->willXxx()` calls above & below have argument specified.


everzet:

PR #337, opened 2017-06-01
Allow PHP5 keywords methods generation on PHP7.

bycosta:
This is the fix of the PR #309 as it's not being updated.

### DIFF
--- a/spec/Prophecy/Doubler/ClassPatch/KeywordPatchSpec.php
+++ b/spec/Prophecy/Doubler/ClassPatch/KeywordPatchSpec.php
@@ -2,6 +2,7 @@
 
 namespace spec\Prophecy\Doubler\ClassPatch;
 
+use PhpSpec\Exception\Example\SkippingException;
 use PhpSpec\ObjectBehavior;
 use Prophecy\Argument;
 use Prophecy\Doubler\Generator\Node\ClassNode;
@@ -25,6 +26,8 @@ class KeywordPatchSpec extends ObjectBehavior
         MethodNode $method2,
         MethodNode $method3
     ) {
+        $this->skipIfPhp7();
+
         $node->removeMethod('eval')->shouldBeCalled();
         $node->removeMethod('echo')->shouldBeCalled();
 
@@ -39,5 +42,12 @@ class KeywordPatchSpec extends ObjectBehavior
         ));
 
         $this->apply($node);
+    }
+
+    private function skipIfPhp7()
+    {
+        if (\PHP_VERSION_ID >= 70000) {
+            throw new SkippingException('Reserved keywords list in PHP 7 does not include most of PHP 5.6 keywords');
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


spec/Prophecy/Doubler/ClassPatch/KeywordPatchSpec.php
```diff
@@ -25,6 +26,8 @@ function it_will_remove_echo_and_eval_methods(
         MethodNode $method2,
         MethodNode $method3
     ) {
+        $this->skipIfPhp7();
```
Taluu: instead of a function call, I would directly put the if + throw here (more readable, more maintainable, one less function call, ...)

stof: totally agreed

stof: We still need a non-skipped test covering the case of `__halt_compiler` (and probably a test covering that `echo` *is* allowed on PHP 7)

bycosta:
@Taluu @stof Done!

bycosta:
@Taluu @stof Can you please review this again? Thanks!

bycosta:
@stof ping!

bycosta:
ping @stof

everzet:
Thank you for contribution!

elkaadka:
Why on earth would you merge this PR,
Also the tests makes no sense
And why would you think __halt_compiler can not be used in php7 as a method name ? 

do some tests before coding for god sake

http://sandbox.onlinephpfunctions.com/code/1d5df3c7a5f4a08c2b900a46f7130616e1342139

avant1:
@elkaadka `__halt_compiler` cannot be used as method name in PHP 7.0.6+: https://3v4l.org/lTBqJ

xurumelous:
The art of erasing a comment when you were wrong and arrogant as *#^&.

elkaadka:
Well your answer only proved one thing.
Your Incompetence.. i deleted my comment cause i didn't want my name to show up in such a poorly written package.
__halt_compiler World for php 7 to 7.0.6 but let's ban it anyway cause we're lazy and f**** the poeple who might be needing it

sstok:
```
Well your answer only proved one thing.
Your Incompetence.. i deleted my comment cause i didn't want my name to show up in such a poorly written package. __halt_compiler World for php 7 to 7.0.6 but let's ban it anyway cause we're lazy and f**** the poeple who might be needing it
```

First of why would you need to call your function/method `__halt_compiler`? PHP explicitly states that function names starting with double underscore are considered magic and you shouldn't rely on it for your own functions.

Secondly https://3v4l.org/INSrC this only worked for PHP 7.0.0 till PHP 7.0.5, it was properly a bug or something and relying on this is very dangerous.

> i didn't want my name to show up in such a poorly written package

Give me a break... I suggest you tone a little.

PR #341, opened 2017-07-10
check method predictions only once

### LINKED ISSUE #326: Prediction checked second time after test method
A shouldBeCalled prediction that succeeded once is checked another time after the test method in a PHPUnit test. If the prediction checks an object that had been changed in the meantime, it might fail. Here an example:
```
interface TestInterface {
    public function test(\stdClass $obj);
}

class DuplicatePredictionCheckTest extends \PHPUnit_Framework_TestCase
{
    public function test()
    {
        $obj = new \stdClass();
        $obj->foo = 'bar';
        $prophecy = $this->prophesize(TestInterface::class);
        $prophecy->test(\Prophecy\Argument::that(function ($obj) {
            static $calls = 0;
            $calls++;
            if ('bar' !== $obj->foo) {
                throw new \RuntimeException(sprintf('Called %d times', $calls));
            }

            return true;
        }))->shouldBeCalled();
        // prediction is successful
        $prophecy->reveal()->test($obj);
        // prediction will fail later
        $obj->foo = 'baz';
    }
}
```

This will result in a RuntimeException with the message  `Called 2 times` instead of a successful test.
Used versions: Prophecy 1.7.0, PHPUnit 5.7.17

dontub:
Fixes #326.

### DIFF
--- a/spec/Prophecy/Call/CallCenterSpec.php
+++ b/spec/Prophecy/Call/CallCenterSpec.php
@@ -3,10 +3,12 @@
 namespace spec\Prophecy\Call;
 
 use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+use Prophecy\Argument\ArgumentsWildcard;
+use Prophecy\Call\Call;
 use Prophecy\Promise\PromiseInterface;
 use Prophecy\Prophecy\MethodProphecy;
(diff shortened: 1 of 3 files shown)


everzet:
@dontub thanks for contribution!

PR #344, opened 2017-07-12
Make keywords usable as methods names starting PHP 7.0

elkaadka:


src/Prophecy/Doubler/ClassPatch/KeywordPatch.php
```diff
@@ -39,6 +39,10 @@ public function supports(ClassNode $node)
      */
```
elkaadka: __halt_compiler can in fact be used as a function name starting php7
http://sandbox.onlinephpfunctions.com/code/1e794a732c89480a317b12c76952b7e3b5cdd0be

PR #346, opened 2017-07-25
Remove broken requirement of php7

dagnew:
prophecy has a requirement of "doctrine/instantiator": "^1.0.2"; as of  22/07/2017 the dependency package doctrine/instantiator requires php: ^7.1 which breaks backwards compatibility with php 5.x. 
Is it possible to specify a required fixed version of doctrine/instantiator" "1.0.2" instead of "^1.0.2" as the "^1.0.2" package loads 1.1.0   requires php7

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "php":                               "^5.3|^7.0",
         "phpdocumentor/reflection-docblock": "^2.0|^3.0.2",
         "sebastian/comparator":              "^1.1|^2.0",
-        "doctrine/instantiator":             "^1.0.2",
+        "doctrine/instantiator":             "1.0.2",
         "sebastian/recursion-context":       "^1.0|^2.0|^3.0"
     },
 


sstok:
This shouldn't affect you unless you have PHP 7 installed, if your production system and development system differ you can set the Composer platform-version to a specific version.
https://getcomposer.org/doc/06-config.md#platform

If this affect any other package that you use, then those packages need to updated instead. Blocking versions like this would make it impossible for anyone to use a newer version of the instantiator.

ciaranmcnulty:
Yeah this isn't necessary, users on PHP 5 will get 1.0.2 and users on PHP 7 will get a newer version.

Fine so long as we test with the lower versions too

dagnew:
@sstok @ciaranmcnulty  - thanks for the responses guys, appreciated. We're running 5.6 on dev and prod  and this is specified in composer.json, but it still pulling the doctrine/instantiator (1.1.0) package instead of the 1.0.2. If you've any ideas, they'd be welcome :)

ciaranmcnulty:
The instantiator `"require": {"php": "7.1:}` should prevent that

Maybe raise a Composer bug?

dagnew:
@ciaranmcnulty  - good point, thanks :+1:  I'll close this PR.

Ocramius:
Just jumping in to add that version requirement bumps are not a bc break.

PR #350, opened 2017-08-04
[WIP] Allow nullable typehinted arguments [PHP 7.1]

ruudk:
In a project we are using variadic arguments. In PHP 7.1 we now have the power of nullable type hints and return types. You can even do nullable typehinted variadic arguments like this:
```php
function send(?Message ...$messages) {
    foreach ($messages as $message) {
        if (null === $message) {
            continue;
        }

        // send 
    }
}

send(new Message(), null, new Message());
```

This works fine. But it doesn't work in PHPSpec:
```
warning: Declaration of Double\MailService\P2306::addMessageAndSend(Message ...$messages): void should be compatible with MailService::addMessageAndSend(?Message ...$messages): void in vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(49)
```

I already solved this issue in [PHPUnit's Mock Objects ](https://github.com/sebastianbergmann/phpunit-mock-objects/pull/372) and that was really easy. But it's a lot harder in this project.

Could somebody point me in the right direction here?

The issue I have currently is that this source:
```php
public function methodWithArgs(bool $arg_1 = true, ?bool $arg_2, ?bool $arg_3 = true, ?bool $arg_4 = null)
```

becomes this:
```php
public  function methodWithArgs(bool $arg_1 = true, ?bool $arg_2 = NULL, ?bool $arg_3 = true, ?bool $arg_4 = NULL)
```

The second argument should not have ` = NULL`

### DIFF
--- a/fixtures/WithNullableArguments.php
+++ b/fixtures/WithNullableArguments.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace Fixtures\Prophecy;
+
+class WithNullableArguments
+{
+    public function methodWithArgs(bool $arg_1 = true, ?bool $arg_2, ?bool $arg_3 = true, ?bool $arg_4 = null)
+    {
+    }
+}
--- a/src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
+++ b/src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
@@ -135,7 +135,7 @@ class ClassCodeGenerator
 
             $php .= '$'.$argument->getName();
 
-            if ($argument->isOptional() && !$argument->isVariadic()) {
+            if ($argument->hasDefault()) {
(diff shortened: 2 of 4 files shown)


Taluu:
I think I have a similar problem ; I have an interface that takes the following signature :

```php
<?php
interface MyInterface
{
    public function myMethod(?string $foo = 'bar');
}
```

which then becomes

```php
<?php
public function myMethod(string $foo = 'bar') { }
```

So it seems that if an argument is "nullable", and has a default value which is not null, it removes the nullable...

ruudk:
@everzet Could you maybe point me in the right direction so that I can fix this PR? 🙏

everzet:
Tests are failing :)

ruudk:
@everzet Removed.

PR #354, opened 2017-08-12
Allow reflection-docblock v4

GrahamCampbell:


Taluu:
How about using the real or notation of composer (`||` instead of `|`) while you're at it ?

GrahamCampbell:
I prefer the single bar notation. :P (I don't think the double one is a replacement, just an alternative?)

Taluu:
Actually both are indeed supported, but in [the docs](https://getcomposer.org/doc/articles/versions.md#version-range) there's no mention of `|`, which could mean IMO that this is "kind of deprecated" / "hidden".

derrabus:
The single pipe is consistent with the rest of the file. It's a deprecated notation, but it will probably be supported forever. 😉

jaapio:
It would be very nice if this can be merged soon. It is blocking new developments in phpdocumentor itself now :-(

GrahamCampbell:
Ping @everzet. ;)

everzet:
@GrahamCampbell thanks for contribution

PR #367, opened 2017-11-01
[bug] CallbackPromise/CallbackPrediction Try to bind to static lambda.

SpacePossum:
Hi,

When doing something like:

```php
$fixer = $this->prophesize();
$fixer->willImplement(\PhpCsFixer\Fixer\DefinedFixerInterface::class);
$fixer->fix(
    Argument::type(\SplFileInfo::class),
    Argument::type(\PhpCsFixer\Tokenizer\Tokens::class)
)->will(static function (array $arguments) use (&$things) {
});
```

it will fail with a message like;

```
[03:16 PM]-[possum@zoo1]-[/home/possum/work/prophecy]-[git master] 
$ ./vendor/bin/phpunit tests/Promise/CallbackPromiseTest.php
PHPUnit 5.7.23 by Sebastian Bergmann and contributors.

E                                                                   1 / 1 (100%)

Time: 49 ms, Memory: 4.00MB

There was 1 error:

1) Tests\Prophecy\Promise\CallbackPromiseTest::testCallbackPromise
Cannot bind an instance to a static closure

/home/possum/work/prophecy/tests/Promise/CallbackPromiseTest.php:22

ERRORS!
```

This is because `CallbackPromise` (and `CallbackPrediction`) try to bind to a static lambda,
which is not allowed.

For ref.: 
- http://php.net/manual/en/closure.bindto.php
- http://php.net/manual/en/functions.anonymous.php#functions.anonymous-functions.static (`bindTo` note)
- real example demonstrating the issue; https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3187/files#diff-16ba030c9b715a9ab3427b5ba9c3ea25R241

### DIFF
--- a/tests/Promise/CallbackPromiseTest.php
+++ b/tests/Promise/CallbackPromiseTest.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Tests\Prophecy\Promise;
+
+/**
+ * @internal
+ */
+final class CallbackPromiseTest extends \PHPUnit_Framework_TestCase
+{
+    public function testCallbackPromise()
+    {
+        $fixer = $this->prophesize('Tests\Prophecy\Promise\CallbackPromiseTestDummy');
+        $fixer->getName()->will(static function (array $arguments) use (&$things) {
+            return $things
+                ? 'yes'
+                : 'no'
+            ;
+        });
+
+        $this->assertSame('no', $fixer->reveal()->getName());
+    }
+}
+
+class CallbackPromiseTestDummy
+{
+    public function getName() {
+        return 'dummy';
+    }
+}
--- a/src/Prophecy/Prediction/CallbackPrediction.php
+++ b/src/Prophecy/Prediction/CallbackPrediction.php
@@ -56,7 +56,7 @@ class CallbackPrediction implements PredictionInterface
     {
         $callback = $this->callback;
 
-        if ($callback instanceof Closure && method_exists('Closure', 'bind')) {
+        if ($callback instanceof Closure && method_exists('Closure', 'bind') && null === (new \ReflectionFunction($callback))->getClosureScopeClass()) {
             $callback = Closure::bind($callback, $object);
         }
 
--- a/src/Prophecy/Promise/CallbackPromise.php
+++ b/src/Prophecy/Promise/CallbackPromise.php
@@ -57,7 +57,7 @@ class CallbackPromise implements PromiseInterface
     {
         $callback = $this->callback;
 
-        if ($callback instanceof Closure && method_exists('Closure', 'bind')) {
+        if ($callback instanceof Closure && method_exists('Closure', 'bind') && null === (new \ReflectionFunction($callback))->getClosureScopeClass()) {
             $callback = Closure::bind($callback, $object);
         }
 


tests/Promise/CallbackPromiseTest.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Tests\Prophecy\Promise;
+
+/**
+ * @internal
+ */
+final class CallbackPromiseTest extends \PHPUnit_Framework_TestCase
```
keradus: shouldn't we add the same for `CallbackPrediction` ?

SpacePossum: Maybe best :+1: 
TBH I'm waiting on instructions of the maintainers/members here about how to proceed.

keradus: adding missing test would be state of art of what you want to fix, giving repo owners way to merge it when they see it for first time :)

SpacePossum: I've a feeling they don't want a utest like this, but in the form of a `fixture` (https://github.com/phpspec/prophecy/tree/master/fixtures). However I don't understand these yet, no do I know if they really want this.

stof: testing this case should be done in the spec of the class. We use PHPUnit for integration tests. Covering a single unit is done in our PhpSpec suite instead, and this case can be perfectly covered there instead of using a functional test here (the test you wrote is a functional test).

and yes, you need to have a similar coverage for CallbackPrediction, to ensure it does not break in the future.

SpacePossum:
Closing, cause after more testing I found no reliable way to determine at runtime if a Closure bind can be done.
For ref.: https://3v4l.org/gZOfH

keradus:
@SpacePossum what about replace the check with trying to bind and catch error (sadly, error, not exception)

SpacePossum:
catching the error would require setting up an error handler at that point, it is going to be expensive...
maybe this should be fixed within PHP itself, adding a getter to the Callable object?

keradus:
i wouldn't expect that...

@stof , you know now what the issue is, what would you advise ?

deguif:
Just see this issue as I'm encountering the same error.
Couldn't we just silently ignore the error, by adding `@` before the `Closure::bind`?

SpacePossum:
I think best would be to ask for a `is_bindable()` function in the global namespace by PHP itself in PHP7.5 or 8, not sure if anyone has experience with this?

For the VM it seems easy to test this, however in user land it is a mess.

deguif:
Yeah for sure, implementing this on PHP side would be nice.
In the mean time, do you think `@Closure::bind` could be a workaround?

I'm really struggling with this issue when enabling `static_lambda` on `php-cs-fixer` as you ;)

SpacePossum:
I gave up on this issue (and the usage of the `static_lambda` rule in some projects :( ) when I couldn't find a reliable solution that would not involve dealing with the error handler. There might be a good solution still.
Secondly writing the required (unit) tests for this change is not something I can invest time in.

Would really like to see this resolved, I just cannot commit to it as I have no time for it.

PR #369, opened 2017-11-07
Ensure the classname is a string, not an object

greg0ire:


src/Prophecy/Doubler/ClassPatch/ProphecySubjectPatch.php
```diff
@@ -83,7 +83,7 @@ public function apply(ClassNode $node)
         $__call->setCode(<<<PHP
 throw new \Prophecy\Exception\Doubler\MethodNotFoundException(
     sprintf('Method `%s::%s()` not found.', get_class(\$this), func_get_arg(0)),
-    \$this->getProphecy(), func_get_arg(0)
+    \\get_class(\$this->getProphecy()), func_get_arg(0)
```
keradus: no tests for it ? :(

greg0ire: It it didn't fail before, I guess there are no tests :P

keradus: that's the thing.
then, one will break it again in future, as there are no tests :(

keradus: i'm not saying to add tests for everything here, but for that one thing you just fix

greg0ire: I'll try to have a look at it, but before, do you think it is the right fix? Or should it actually be `$this->getProphecy()->getMockedClassNameOrWhateverTheCorrectMethodIs()`?

keradus: tbh, dunno, i didn't analysed it :(

greg0ire: Right now, with the fix, the error message I get is about a method not existing on the double (instead of on the original class).

avant1:
As I understand original issue is related to PHPSpec: phpspec/phpspec#1147
It also contains reproducible scenario.

src/Prophecy/Doubler/ClassPatch/ProphecySubjectPatch.php
stof: Passing the class of the prophecy does not make much sense. This will always be ObjectProphecy, and this is not the class on which the method is undefined.
We should rather use the name of the mocked class (using `get_class($this)` as done in the sprintf)

greg0ire: @stof it's not always `ObjectProphecy`, the only thing I witnessed here is `Double\My\Original\Class\P2`

greg0ire: But yeah, let's use `get_class($this)`

greg0ire: Wait actually, I'm not sure anymore, what I witnessed must be `get_class($this)`, and you were problaby right about `ObjectProphecy`. That means using `get_class($this)` in the message is probably not best?

stof: Well, saying that the method does not exist on the ObjectProphecy does not make any sense. The call is not performed on it at all (the ObjectProphecy holds the configuration).

The message is using `get_class($this)`, which is much better (even better would be to account for the actual class or interface being doubled, but we can have multiple ones, so this becomes hard)

greg0ire: Ok so let's stick with `get_class($this)`

arendjantetteroo:
@avant1 this one is triggered on line 95 of the CollaboratorMethodNotFoundlIstener  -> getMethodNotFoundException

Where 1147 is on line 99, for getDoubledInterface. 

The messages look the same, but this fix does not fix 1147.

greg0ire:
Maybe there is a similar fix to be done? Can you use github permalinks, so that we can see the code you are referring to?

arendjantetteroo:
https://github.com/phpspec/phpspec/blob/master/src/PhpSpec/Listener/CollaboratorMethodNotFoundListener.php#L95-L101

greg0ire:
That's not a permalink. To have one, click on the `…`. Here is one: https://github.com/phpspec/phpspec/blob/30f329cf253e581dac4410d5dbbd8b778b180028/src/PhpSpec/Listener/CollaboratorMethodNotFoundListener.php#L95-L101

arendjantetteroo:
Ah sorry, thanks for pointing it out :)

greg0ire:
No worries, it doesn't work anyway unless the file is in the same repo :P

greg0ire:
Looking at your file, this is exactly where I got my error message too. Type hinting could help getting it earlier here.

arendjantetteroo:
I just realised i was changing it in the wrong vendor directory, had 2 projects open. It works now. Now get a proper error message. Sorry :(

arendjantetteroo:
18  - it can be segmented
      method `Double\Example\Segment\P1::forBusinessCycle()` not found.

greg0ire:
Looks great.

arendjantetteroo:
So thanks, this indeed solves 1147

spec/Prophecy/Doubler/ClassPatch/ProphecySubjectPatchSpec.php
```diff
@@ -37,6 +37,21 @@ function it_forces_class_to_implement_ProphecySubjectInterface(ClassNode $node)
...
+        $node->addInterface('Prophecy\Prophecy\ProphecySubjectInterface')->shouldBeCalled();
+
+        $node->addProperty('objectProphecy', 'private')->willReturn(null);
+        $node->getMethods()->willReturn(array());
+        $node->hasMethod(Argument::any())->willReturn(false);
+        $node->addMethod(Argument::type('Prophecy\Doubler\Generator\Node\MethodNode'))->willReturn(null);
+        $this->apply($node);
+        $node->addMethod(Argument::which('getName', '__call'))->shouldBeCalled();
+        $node->addMethod(Argument::that(function ($method) {
+            return strpos($method->getCode(), 'not found') !== false;
```
greg0ire: @keradus I tried adding some coverage, but this test would not fail w/o my commit, and I don't know how to test it: I tried using `eval` on the code, but `func_get_args` has no meaning outside a method. It feels overkill.

keradus: if we cannot prove the bug in automated way, how could we know there is a bug in first place?
how one gonna protect that it works over time and other PRs didn't reintroduce this issue?

greg0ire: @keradus I'd say with an integration test, maybe?

keradus: for me any automated test would be good :+1:

greg0ire: First I'll see if I can make the tests fail by simply adding some type hints.

greg0ire: Oh no I can't php 5 is still supported.

greg0ire: I'm going to give `eval` another try

greg0ire:
What do you think of this @keradus ?

keradus:
I think new tests pass without the fix, still

keradus:
does new tests fails without the fix?
without the fix, `it_adds_a_magic_call_implementation_that_instantiates_method_not_found_properly` is freezing at my machine :(

greg0ire:
It brutally stops with 255 on mine...

greg0ire:
It's because the `eval()` execute codes that converts an object to string I think. If I use "echo 'test';" as `sprintf`, it works just fine. That's how broken the current code is. Anyway, is it important if it crashes instead of failing? Since the exit code is 255?

keradus:
that was my point, it was not crashing, it was freezing, exit code was never provided ;)

greg0ire:
Should that happen on Travis, it would be red too, there is a timeout.

greg0ire:
ping @keradus is that good enough?

keradus:
for me it's acceptable, as I assume timeout issue is my local env issue.
yet, ultimate decision is to be done by repo maintainers/owners

greg0ire:
I spoke with @ciaranmcnulty , this isn't the right fix. This bug was introduced when adding type hinting here: https://github.com/phpspec/phpspec/blob/8d3a2dbc8eb8c9b00fc832d9adc825a15135fdcd/src/PhpSpec/Listener/CollaboratorMethodNotFoundListener.php#L116

PR #373, opened 2017-11-20
Added support to Symfony Finder SplFileInfo

gnugat:
In Symfony 4, SplFileInfo provided by the Finder component requires its
constructor arguments to be strings
(see
https://github.com/symfony/symfony/commit/aaf2265203cfbefa006f0358077d8ade790ad203#diff-1c5661d50aeb4ac4fc0f6c0566fc2196L29).

This brings a few issues in phpspec (see
https://github.com/phpspec/phpspec/pull/1166#issuecomment-345616165),
this patch aims at fixing those issues.

Hopefully this patch isn't to farfetched, I assumed it was OK as there
are already a couple of patches for SplFileInfo.

### DIFF
--- a/src/Prophecy/Doubler/ClassPatch/SplFileInfoPatch.php
+++ b/src/Prophecy/Doubler/ClassPatch/SplFileInfoPatch.php
@@ -34,7 +34,6 @@ class SplFileInfoPatch implements ClassPatchInterface
         if (null === $node->getParentClass()) {
             return false;
         }
-
         return 'SplFileInfo' === $node->getParentClass()
             || is_subclass_of($node->getParentClass(), 'SplFileInfo')
         ;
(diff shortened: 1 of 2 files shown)


src/Prophecy/Doubler/ClassPatch/SplFileInfoPatch.php
```diff
@@ -34,7 +34,6 @@ public function supports(ClassNode $node)
         if (null === $node->getParentClass()) {
             return false;
         }
-
```
sroze: Not sure why this line removal?

src/Prophecy/Doubler/ClassPatch/SplFileInfoPatch.php
```diff
@@ -103,4 +109,15 @@ private function nodeIsSplFileObject(ClassNode $node)
...
+
+    /**
+     * @param ClassNode $node
+     * @return boolean
+     */
+    private function nodeIsSymfonySplFileInfo(ClassNode $node)
+    {
+        $parent = $node->getParentClass();
+
+        return 'Symfony\\Component\\Finder\\SplFileInfo' === $parent;
```
ciaranmcnulty: You can use `::class` here, and I don't think it's worth having the `$parent` variable

stof: nope. composer.json still says 5.3

PR #379, opened 2017-12-15
Don't truncate strings by default

### LINKED ISSUE #340: Why does Prophecy truncate expected arguments on comparison failure ?
Let's consider this use case : 

```php
    public function testItAddImageSizesWhenNotProvided()
    {
        $initialPostContentImage = '<img src="/tmp/imaged0eTef" />';
        $expectedPostContentImage =  '<img src="/tmp/imaged0eTef" width="600" height="400" />';

        $post = $this->prophesize(Post::class);
        $post->getContent()->willReturn($initialPostContentImage);
        $post->replaceContent($expectedPostContentImage)->shouldBeCalledTimes(1);

        $calculateImagesSizes = new CalculateImageSizes();
        $calculateImagesSizes->filter($post->reveal());
    }
```

yield 

```
Method call:
  - replaceContent("<img src="/tmp/imaged0eTef" width="600" height="4"...)
on Double\Europe1\FrontBundle\Client\Scribble\Model\Post\P1 was not expected, expected calls were:
  - getContent()
  - replaceContent(exact("<img src="/tmp/imaged0eTef" width="600" height="4"...))
```

Because of the truncated string, I could not see that the actual string passed to the `replaceContent()` was ``.

I their a reason why strings gets truncated in [`Prophecy/Util/StringUtil.php`](https://github.com/phpspec/prophecy/blob/d883566b83ae601a50b38b249e92450dc57cf875/src/Prophecy/Util/StringUtil.php#L58) ?

dunglas:
Closes #340.

This behavior is very annoying:

```
  - getCollection("ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity"...)
on Double\CollectionDataProviderInterface\P16 was not expected, expected calls were:
  - getCollection(exact("ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity"...))
```

With this change:

```
Prophecy\Exception\Call\UnexpectedCallException: Method call:
  - getCollection("ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\Dummy")
on Double\CollectionDataProviderInterface\P16 was not expected, expected calls were:
  - getCollection(exact("ApiPlatform\Core\Tests\Fixtures\TestBundle\Entity\RelatedDummy"))
```

### DIFF
--- a/src/Prophecy/Util/StringUtil.php
+++ b/src/Prophecy/Util/StringUtil.php
@@ -20,6 +20,16 @@ use Prophecy\Call\Call;
  */
 class StringUtil
 {
+    private $verbose;
+
+    /**
+     * @param bool $verbose
+     */
+    public function __construct($verbose = true)
+    {
+        $this->verbose = $verbose;
+    }
+
     /**
      * Stringifies any provided value.
      *
@@ -54,7 +64,7 @@ class StringUtil
         if (is_string($value)) {
             $str = sprintf('"%s"', str_replace("\n", '\\n', $value));
 
-            if (50 <= strlen($str)) {
+            if (!$this->verbose && 50 <= strlen($str)) {
                 return substr($str, 0, 50).'"...';
             }
 


stof:
Do we actually need this verbose property, if nothing in Prophecy ever use it ?

dunglas:
It was to keep a way to enable the old behavior. I'm fine with just dropping the truncating as it hurts more than it helps.

stof:
can you describe me how people can restore the old behavior in Prophecy then ? Where would they turn off verbose mode ? As far as I remember, the StringUtil is instantiated internally all the time.

dunglas:
```php
$prophet = new Prophet(null, null, new StringUtil(false));
```

dunglas:
@stof, do you think this one can be merged?

PR #382, opened 2017-12-21
[CS] Combine consecutives issets

carusogabriel:
`isset` allows us to pass multiple args to check.

[`PHP-CS-Fixer`](https://github.com/FriendsOfPHP/PHP-CS-Fixer) helped me on that.

### DIFF
--- a/src/Prophecy/Call/CallCenter.php
+++ b/src/Prophecy/Call/CallCenter.php
@@ -68,7 +68,7 @@ class CallCenter
         }
 
         $file = $line = null;
-        if (isset($backtrace[2]) && isset($backtrace[2]['file'])) {
+        if (isset($backtrace[2], $backtrace[2]['file'])) {
             $file = $backtrace[2]['file'];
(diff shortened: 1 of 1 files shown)


stof:
I tend to find the 2 separate calls more readable here (not having to remember whether it is a AND or a OR. What is the benefit here ?

carusogabriel:
No benefit in terms of speed, improvement or so. Is up to you :+1:

src/Prophecy/Call/CallCenter.php
```diff
@@ -68,7 +68,7 @@ public function makeCall(ObjectProphecy $prophecy, $methodName, array $arguments
         }
 
         $file = $line = null;
-        if (isset($backtrace[2]) && isset($backtrace[2]['file'])) {
+        if (isset($backtrace[2], $backtrace[2]['file'])) {
```
Taluu: even better : `isset($backtrace[2]['file'])` as for one unset key (e.g 2), will not trigger errors and return false.

PR #389, opened 2018-02-11
Generate proper object return type

greg0ire:
Generates `\object` on php < 7.2, `object` otherwise, so madmen who want to call their classes object ( or [bool](https://github.com/ruflin/Elastica/blob/2.x/lib/Elastica/Query/Bool.php)) can.

### DIFF
--- a/src/Prophecy/Doubler/Generator/Node/MethodNode.php
+++ b/src/Prophecy/Doubler/Generator/Node/MethodNode.php
@@ -145,11 +145,9 @@ class MethodNode
                 // Fall-through to default case for PHP < 7.2
 
             default:
-                if ($this->typeHintReference->isBuiltInReturnTypeHint($type)) {
-                    $this->returnType = $type;
-                } else {
-                    $this->returnType = '\\' . ltrim($type, '\\');
-                }
+                $this->returnType = $this->typeHintReference->isBuiltInReturnTypeHint($type) ?
+                    $type :
+                    '\\' . ltrim($type, '\\');
         }
     }
 
--- a/src/Prophecy/Doubler/Generator/Node/MethodNode.php
+++ b/src/Prophecy/Doubler/Generator/Node/MethodNode.php
@@ -137,6 +137,13 @@ class MethodNode
                 $type = 'int';
                 // intentional fall through
 
+            case 'object':
+                if (version_compare(PHP_VERSION, '7.2', '>=')) {
+                    $this->returnType = $type;
+                    break;
+                }
+                // Fall-through to default case for PHP < 7.2
+
             default:
                 if ($this->typeHintReference->isBuiltInReturnTypeHint($type)) {
                     $this->returnType = $type;
--- a/src/Prophecy/Doubler/Generator/Node/MethodNode.php
+++ b/src/Prophecy/Doubler/Generator/Node/MethodNode.php
@@ -11,6 +11,7 @@
 
 namespace Prophecy\Doubler\Generator\Node;
 
+use Prophecy\Doubler\Generator\TypeHintReference;
 use Prophecy\Exception\InvalidArgumentException;
 
 /**
@@ -34,13 +35,19 @@ class MethodNode
     private $arguments = array();
 
     /**
+     * @var TypeHintReference
+     */
+    private $typeHintReference;
+
+    /**
      * @param string $name
      * @param string $code
      */
-    public function __construct($name, $code = null)
+    public function __construct($name, $code = null, TypeHintReference $typeHintReference = null)
     {
         $this->name = $name;
         $this->code = $code;
+        $this->typeHintReference = $typeHintReference ?: new TypeHintReference();
     }
 
     public function getVisibility()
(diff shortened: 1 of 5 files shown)


greg0ire:
> Breaks void return type.

@Majkl578 fixed, please review again

src/Prophecy/Doubler/Generator/Node/MethodNode.php
```diff
@@ -117,32 +124,32 @@ public function setReturnType($type = null)
...
             case 'boolean':
-                $this->returnType = 'bool';
-                break;
+                $type = 'bool';
+                // intentional fall through
 
             case 'integer':
-                $this->returnType = 'int';
-                break;
+                $type = 'int';
```
Majkl578: This will always overwrite `$type` if they were double/real/boolean, right? Looks fishy.

src/Prophecy/Doubler/Generator/Node/MethodNode.php
```diff
@@ -117,32 +124,32 @@ public function setReturnType($type = null)
...
+                    $this->returnType = $type;
+                    break;
+                }
+                // Fall-through to default case for PHP < 7.2
 
             default:
-                $this->returnType = '\\' . ltrim($type, '\\');
+                if ($this->typeHintReference->isBuiltInReturnTypeHint($type)) {
+                    $this->returnType = $type;
+                } else {
```
Majkl578: Any chance to avoid else, i.e. by ternary?

src/Prophecy/Doubler/Generator/Node/MethodNode.php
greg0ire: Maybe, but it seems to work: https://3v4l.org/KrZmm

olvlvl: Because cases fall through, their `type` will be overwritten by subsequent cases: https://3v4l.org/LtCg4, `real` will end up being `int`. Better move the logic in _default_ outside of the _switch_ statement and have each _case_ use _break_.

src/Prophecy/Doubler/Generator/Node/MethodNode.php
```diff
@@ -117,32 +124,30 @@ public function setReturnType($type = null)
...
             case 'integer':
-                $this->returnType = 'int';
-                break;
+                $type = 'int';
+                // intentional fall through
+
+            case 'object':
+                if (version_compare(PHP_VERSION, '7.2', '>=')) {
+                    $this->returnType = $type;
+                    break;
```
olvlvl: We could replace `break` by `return`.

src/Prophecy/Doubler/Generator/Node/MethodNode.php
greg0ire: Oh wow looks like `switch` does not work like I think it does… Once a case is matched, the other are just ignored?

greg0ire: It's explained in detail in the docs… wtf

src/Prophecy/Doubler/Generator/Node/MethodNode.php
greg0ire: There's no need for this case :P

src/Prophecy/Doubler/Generator/Node/MethodNode.php
```diff
@@ -119,36 +119,22 @@ public function hasReturnType()
...
+        $typeMap = [
+            'double' => 'float',
+            'real' => 'float',
+            'boolean' => 'bool',
+            'integer' => 'int',
+        ];
+        if (isset($typeMap[$type])) {
+            $type = $typeMap[$type];
+        }
+        $this->returnType = $this->typeHintReference->isBuiltInReturnTypeHint($type) ?
```
olvlvl: You forgot the special case with `object`.

src/Prophecy/Doubler/Generator/Node/MethodNode.php
```diff
@@ -119,36 +119,22 @@ public function hasReturnType()
...
-
-            default:
-                $this->returnType = $this->typeHintReference->isBuiltInReturnTypeHint($type) ?
-                    $type :
-                    '\\' . ltrim($type, '\\');
+        if ($type === '') {
+            $this->returnType = null;
+            return;
         }
+        $typeMap = [
```
olvlvl: You need to use `array(` here, PHP 5.3…

greg0ire: Argh

olvlvl:
@ciaranmcnulty Can we get this merged? 🙏

ciaranmcnulty:
Looks good

Majkl578:
@ciaranmcnulty Can you please release new version with this change included? It is blocking progress on 7.2 upgrade currently. Thanks. :tada:

ciaranmcnulty:
I believe it was in 1.7.5

Majkl578:
Ah right, you changed the tag naming pattern as of 1.7.3 so v1.7.2 is on top, I got confused by that. :) Thanks.

ciaranmcnulty:
Ah crap

PR #391, opened 2018-02-11
Add shouldBeCalledOnce()

olvlvl:
Hi,

This PR allows to write `->shouldBeCalledOnce()` instead of `->shouldBeCalledTimes(1)`. I noticed that most of the times we are using `->shouldBeCalled()` because `shouldBeCalledTimes()` is cumbersome, maybe this could help writing more accurate tests.

### DIFF
--- a/src/Prophecy/Prophecy/MethodProphecy.php
+++ b/src/Prophecy/Prophecy/MethodProphecy.php
@@ -279,6 +279,18 @@ class MethodProphecy
     }
 
     /**
+     * Sets call times prediction to the prophecy.
+     *
+     * @see \Prophecy\Prediction\CallTimesPrediction
+     *
+     * @return $this
+     */
+    public function shouldBeCalledOnce()
+    {
+        return $this->shouldBeCalledTimes(1);
+    }
+
+    /**
      * Checks provided prediction immediately.
      *
      * @param callable|Prediction\PredictionInterface $prediction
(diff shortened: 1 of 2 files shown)


src/Prophecy/Prophecy/MethodProphecy.php
```diff
@@ -372,6 +384,30 @@ public function shouldHaveBeenCalledTimes($count)
...
+     * Checks call times prediction.
+     *
+     * @see \Prophecy\Prediction\CallTimesPrediction
+     *
+     * @return $this
+     */
+    public function shouldHaveBeenCalledTwice()
+    {
+        return $this->shouldHaveBeenCalledTimes(2);
+    }
```
Taluu: only `Once`, right ?

src/Prophecy/Prophecy/MethodProphecy.php
```diff
@@ -372,6 +384,30 @@ public function shouldHaveBeenCalledTimes($count)
...
+
+    /**
+     * Checks call times prediction.
+     *
+     * @see \Prophecy\Prediction\CallTimesPrediction
+     *
+     * @return $this
+     */
+    public function shouldHaveBeenCalledTwice()
```
stof: I don't think we need this one

olvlvl:
Sorry, I forgot this one. Fixed now.

ciaranmcnulty:
Thanks @olvlvl

PR #395, opened 2018-03-16
Added mustReturn() method to MethodProphecy

horak-m:
Method combines willReturn() call with shouldBeCalled() which is frequently used together.

### DIFF
--- a/spec/Prophecy/Prophecy/MethodProphecySpec.php
+++ b/spec/Prophecy/Prophecy/MethodProphecySpec.php
@@ -78,6 +78,17 @@ class MethodProphecySpec extends ObjectBehavior
         $this->will($promise);
     }
 
+    function it_adds_ReturnPromise_and_CallPrediction_during_mustReturn_call($objectProphecy)
+    {
+        $objectProphecy->addMethodProphecy($this)->willReturn(null);
+
+        $this->mustReturn(42);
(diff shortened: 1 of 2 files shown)


Taluu:
Not really sure if it's semantically correct (`mustReturn`). I don't think this will be interpreted as "will return and should be called at least once". So IMO it's fine like this.

Deuchnord:
Why not calling it `shouldBeCalledAndReturn()`?

stof:
I'm not convinced this is needed (and I'm not sure about the "frequently used together". I think I almost never used them together).

meridius:
Well, @stof, we use them together in our team almost every time. It is safer than just `willReturn()` alone because added `shouldBeCalled()` ensures that object not only returns expected result if asked, but also makes sure that the call on it is made. Thus preventing any unnecessary promises.

Deuchnord:
I use `shouldBeCalled()` and `willReturn()` together frequently too, but I'm not sure creating one new method doing both of them is really useful… After all, we can already combine them in our tests.

meridius:
Yes, we can combine them manually for each and every promise but using that alias is so much easier.

ciaranmcnulty:
I agree with @stof, I don't want to encourage this and don't think it's a problem not having it

PR #396, opened 2018-03-24
Fix DisableConstructorPatch breaking already-optional arguments

rakelley:
Currently, the `DisableConstructorPatch` always redefines all constructor arguments as optional with a default value of null, regardless of the argument's existing definition.  This is problematic in cases where the argument is already optional, is not nullable, and the original constructor is invoked.

Example:
`Symfony\Component\HttpFoundation\File\UploadedFile` has a signature of:
```php
__construct(string $path, string $originalName, string $mimeType = null, int $size = null, int $error = null, bool $test = false)
```

Providing all arguments works:
```php
$fileProph = $this->prophesize(UploadedFile::class);
$fileProph->willBeConstructedWith([__FILE__, 'example.txt', 'text/plain', 100, \UPLOAD_ERR_OK, false]);
$fileProph->reveal();
```

I expect the below to work as well, because only the first two arguments are required.  This does not currently work, because the generated double constructor has `bool $test = NULL`, which is incompatible with the original constructor.
```php
$fileProph = $this->prophesize(UploadedFile::class);
$fileProph->willBeConstructedWith([__FILE__, 'example.txt']);
$fileProph->reveal();
```

I don't believe the current behavior makes sense, optional arguments with typed defaults should have those defaults retained.

### DIFF
--- a/spec/Prophecy/Doubler/ClassPatch/DisableConstructorPatchSpec.php
+++ b/spec/Prophecy/Doubler/ClassPatch/DisableConstructorPatchSpec.php
@@ -35,8 +35,27 @@ class DisableConstructorPatchSpec extends ObjectBehavior
         $class->getMethod('__construct')->willReturn($method);
         $method->getArguments()->willReturn(array($arg1, $arg2));
 
+        $arg1->isOptional()->willReturn(false);
         $arg1->setDefault(null)->shouldBeCalled();
+        $arg2->isOptional()->willReturn(false);
         $arg2->setDefault(null)->shouldBeCalled();
+
+        $method->setCode(Argument::type('string'))->shouldBeCalled();
+
+        $this->apply($class);
+    }
+
+    function it_does_not_redefine_already_optional_arguments(
+        ClassNode $class,
+        MethodNode $method,
+        ArgumentNode $arg
+    ) {
+        $class->hasMethod('__construct')->willReturn(true);
+        $class->getMethod('__construct')->willReturn($method);
+        $method->getArguments()->willReturn(array($arg));
+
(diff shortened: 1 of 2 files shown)


jakzal:
Looks like a good improvement to me, but what was the reason why it has originally been done this way? The class comment even says `Makes all constructor arguments optional.`.

If this gets merged, the comment needs to be updated ;)

kiler129:
This is actually a very useful thing, since Symfony now triggers false-positive every time test is ran:

```
Passing a size as 4th argument to the constructor of "Symfony\Component\HttpFoundation\File\UploadedFile" is deprecated since Symfony 4.1.
```

Maybe @ciaranmcnulty  can look at this?

ciaranmcnulty:
Like @jakzal says this was deliberate; and TBH I'm not sure why it was done this way historically so I dunno :-)

kiler129:
It makes sense to make them optional and null-assigned in PHP5 era, but when you use types it just explodes ;)

I see this as a BC break if default behavior changes, but maybe there’s a way to parametrize it or provide different way of achieving proper result?

ciaranmcnulty:
Yep we're still supporting 5.3 and don't have a current plan for BC break / 2.0

ciaranmcnulty:
Coming back to this.. why is the constructor even running in this case?

rakelley:
`Prophecy\Doubler\ClassPatch\SplFileInfoPatch` makes it so that any prophecy for a class that extends `\SplFileInfo` has its real constructor invoked, which includes Symfony's various file helper structures.

ciaranmcnulty:
I'm gong to close for now but please re-open if you have time to address it

PR #398, opened 2018-03-31
Fix #397 php 7.2 void return types issue

### LINKED ISSUE #397: Methods type hinted as void in PHP 7.2 fail
Methods type hinted as void in PHP 7.2 fail to have any expectation of a call when used as a Fake.

## Fake unexpected call

```
class SomeThing {
    public function doAThing(): void
    {}
    public function doOtherThing(): int
    { return random_int(0,10); }
}
```

```
class UsesSomeThing {
    function useIt(SomeThing $something) {
        $something->doOtherThing();
        $something->doAThing();
    }
}
```

Given the classes above one would expect to be able to create the Fakes below:

```
    $something = $this->leaderboardRepository = $this->prophesize(SomeThing::class);
    $this->leaderboardRepository->doOtherThing();
    $this->leaderboardRepository->doAThing();

    $usesSomething = new UsesSomeThing;
    $usesSomething->useIt($something->reveal());
```

The fake for `doOtherThing` does get called correctly but when the call is made to the `doAThing` method there is no expectation so prophecy fails with an `UnexpectedCallException`.

There is a workaround which I'm currently using which is to do the following in the test, but this is a bit misleading as it's no longer really a Fake and is inconsistent with how other Fakes are implemented:
```
    $something = $this->leaderboardRepository = $this->prophesize(SomeThing::class);
    $this->leaderboardRepository->doOtherThing();
    $this->leaderboardRepository->doAThing()->will(function(){}); // NOTE CALL TO will()

    $usesSomething = new UsesSomeThing;
    $usesSomething->useIt($something->reveal());
```

The proposed fix will mean that such workarounds are not needed and void fakes work as expected.

crellbar:
Fix #397 php 7.2 void return types issue by adding a default call to will() as for all other types.

### DIFF
--- a/spec/Prophecy/Prophecy/MethodProphecySpec.php
+++ b/spec/Prophecy/Prophecy/MethodProphecySpec.php
@@ -3,9 +3,11 @@
 namespace spec\Prophecy\Prophecy;
 
 use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
 use Prophecy\Argument\ArgumentsWildcard;
 use Prophecy\Call\Call;
 use Prophecy\Prediction\PredictionInterface;
+use Prophecy\Promise\CallbackPromise;
 use Prophecy\Promise\PromiseInterface;
 use Prophecy\Prophecy\ObjectProphecy;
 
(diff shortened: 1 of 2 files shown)


crellbar:
Merged master now that existing HHVM issues from master have been resolved/removed - showing that nothing in this PR caused failures

jakzal:
Sounds good. Would it be possible to cover this with some tests? ;)

ciaranmcnulty:
Thanks @crellbar!

Issue #400, opened 2018-04-15
Can't check for object identity with Argument::is() if they are weakly equal


jnvsor:
When mocking a method `Argument::is` can't be used to distinguish objects that are weakly identical.

Since identity is the whole point of `Argument::is` over `Argument::exact` for objects, this is rather annoying.

If I had to guess the source of the problem is in the `assertEquals` in [`ObjectProphecy::__call`](https://github.com/phpspec/prophecy/blob/dfd6be44111a7c41c2e884a336cc4f461b3b2401/src/Prophecy/Prophecy/ObjectProphecy.php#L251), since the additional prophecies aren't even registered.

Here's an example:

```php
<?php

use Prophecy\Prophet;
use Prophecy\Argument;

require 'vendor/autoload.php';

$p = new Prophet();

$dt1 = new DateTime();
$dt2 = clone $dt1;
$dt3 = clone $dt1;

class TestClass {
    public function methodCall()
    {
    }
}

$prophecyObject = $p->prophesize('TestClass');

// Argument::exact doesn't work because the properties have the same values
$prophecyObject->methodCall(Argument::exact($dt1))->shouldBeCalledTimes(1);
$prophecyObject->methodCall(Argument::exact($dt2))->shouldBeCalledTimes(1);
$prophecyObject->methodCall(Argument::exact($dt3))->shouldBeCalledTimes(1);

// Argument::is doesn't work because for some reason it flat out refuses to
// store more than one IdenticalValueToken even when called repeatedly
$prophecyObject->methodCall(Argument::is($dt1))->shouldBeCalledTimes(1);
$prophecyObject->methodCall(Argument::is($dt2))->shouldBeCalledTimes(1);
$prophecyObject->methodCall(Argument::is($dt3))->shouldBeCalledTimes(1);

// My current workaround
$prophecyObject->methodCall(
    Argument::that(function ($arg) use ($dt1) {
        return $arg === $dt1;
    })
)->shouldBeCalledTimes(1);
$prophecyObject->methodCall(
    Argument::that(function ($arg) use ($dt2) {
        return $arg === $dt2;
    })
)->shouldBeCalledTimes(1);
$prophecyObject->methodCall(
    Argument::that(function ($arg) use ($dt3) {
        return $arg === $dt3;
    })
)->shouldBeCalledTimes(1);

$dummy = $prophecyObject->reveal();

$dummy->methodCall($dt1);
$dummy->methodCall($dt2);
$dummy->methodCall($dt3);

[...]

PR #401, opened 2018-04-16
IdenticalValueToken: Allow identity checks on objects

jnvsor:
ObjectProphecy::__call prevents duplicate methodProphecies by
calling assertEquals on the prophecies, but identity checks
would need assertSame to work properly for objects. If we
store the spl_object_hash in the IdenticalValueToken we'll
cause inequality and let the token be used multiple times.


jnvsor:
Fixes #400 

I've never used phpspec and I can't think of a way to test this

ciaranmcnulty:
I guess this has been neglected because of lack of tests.

It should suffice to recreate the case where the two methodProphecies are applied and both are evaluated?

PR #402, opened 2018-04-18
Allow sebastian/comparator ^3.0

sebastianbergmann:


ciaranmcnulty:
Thanks! Will do a release - Travis failure seems environmental

PR #403, opened 2018-04-18
Drop HHVM support

ciaranmcnulty:
Travis has updated the HHVM version and yet again it's broken the tests

I don't think we should continue supporting it; many projects have dropped support and Facebook are concentrating on HackLang rather than supporting PHP 7 features

@stof do you have an opinion?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,9 +24,6 @@ matrix:
     - php: 7.2
     - php: 5.3
       dist: precise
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
Spoke to @everzet and he concurs so merging

stof:
I totally agree about this too. I don't want to spend time on fixing HHVM compat when the HHVM team does not care about PHP compat anymore.

PR #406, opened 2018-05-09
Do not add the method to the ClassNode when already added

webimpress:
`addMethod` method has now additional parameter - `force`
when the method defintion in ClassNode should be replaced.

Issue is present only in PHP 7.2, example:
```php
interface MyInterface {
    public function get();
}

class MyClass implements MyInterface {
    public function get($force = null) {};
}
```

then it is not possible to prophesize `MyClass` correctly, because method `get` first will be added with `MyClass` definition and then replaced with `MyInterface` definition, what is wrong.

### DIFF
--- a/fixtures/MethodWithAdditionalParam.php
+++ b/fixtures/MethodWithAdditionalParam.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace Fixtures\Prophecy;
+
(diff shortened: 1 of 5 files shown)


ciaranmcnulty:
Thanks @webimpress

PR #408, opened 2018-06-20
CallCenter: improved formatting for unexpected method call message

meridius:
With the original wording it is not really obvious what the message is saying. I propose the new wording which puts key words first and thus makes the message more readable. 

Old look:

```
1485  - it should do something
      method call:
        - getCollectInStoreBundles()
      on Double\Dory\Checkout\Domain\Entity\Basket\P2488 was not expected, expected calls were:
        - getCollectInStoreProducts()
        - getStore()
        - getTelesalesOperatorId()
```

New look:

```
1485  - it should do something
      unexpected method call on Double\Dory\Checkout\Domain\Entity\Basket\P2488:
        - getCollectInStoreBundles()
      expected calls were:
        - getCollectInStoreProducts()
        - getStore()
        - getTelesalesOperatorId()
```

### DIFF
--- a/src/Prophecy/Call/CallCenter.php
+++ b/src/Prophecy/Call/CallCenter.php
@@ -149,23 +149,66 @@ class CallCenter
                                                    array $arguments)
     {
         $classname = get_class($prophecy->reveal());
-        $argstring = implode(', ', array_map(array($this->util, 'stringify'), $arguments));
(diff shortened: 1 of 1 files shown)


meridius:
Tests are in phpspec/phpspec#1208

ciaranmcnulty:
Thanks @meridius - sorry merging took a while but I had to fix it for PHP 5.3

Issue #410, opened 2018-07-18
Cannot prophesize and interface extending \Throwable


dunglas:
If an exception interface extends `\Throwable` (now a recommended best practice, enforced by phpstan), it's not possible to prophesize it.

Reproducer:

```php
interface MyException extends \Throwable {}

$e = $this->prophesize(MyException::class);
$e->willExtend(\Exception::class);
$e->reveal();
```

(moved from https://github.com/symfony/symfony/issues/27419)

ciaranmcnulty:
What error do you get?

dunglas:
> Prophecy\Exception\Doubler\MethodNotExtendableException: Method `getMessage` is not extendable, so can not be added.

Taluu:
Should be closed @ciaranmcnulty, as you fixed it in #412

ostrolucky:
Yes and included test does not use same thing code posted here does. It does not extend Exception but stdClass. So once again, you have failing test case in this issue, which was never fixed.

ciaranmcnulty:
I’m not clear why the example adds willExtend(Exception) - can you clarify why the test is adding that?

ostrolucky:
Actually, patch explicitly forbids this use case with test case in https://github.com/phpspec/prophecy/blob/3660b2eed90abdf92ac2e9dae9a238d2701bfa77/spec/Prophecy/Doubler/ClassPatch/ThrowablePatchSpec.php#L42-L51

ciaranmcnulty:
No that is excluding concrete Exceptions or Errors (which don’t need patching)

ostrolucky:
> can you clarify why the test is adding that?

PHP5 compatible library can't use `->willImplement('Throwable')` so has to do `->willExtend('Exception')` in case of mocking interface known to be thrown

```php
$mayOrMayNotBeThrowable = rand(0,1) ? 'Throwable' : 'stdClass';
$this->prophesize($mayOrMayNotBeThrowableClass)->willExtend(\Exception::class)->reveal();
```

dunglas:
For those wanting looking for a quick userland fix supporting the use case described  by @ostrolucky (basically, Symfony), this works:

```php
$serializerException = $this->prophesize(ExceptionInterface::class);
if (!is_a(ExceptionInterface::class, \Throwable::class, true)) {
    $serializerException->willExtend(\Exception::class);
}
```

ostrolucky:
[...]

PR #412, opened 2018-07-30
Allow to double Throwable, or an interface that extends Throwable

ciaranmcnulty:
The use case is explained in #410 

When doubling something like the following:

```
interface VendorProblem extends Throwable {}
```

 Prophecy was trying to generate something like:

```
class DoubleOfFoo extends stdClass implements MyTraversable
```

This is not allowed by PHP. To fix, there is a ClassPatch that sets the class being extended to Exception instead of stdClass


src/Prophecy/Doubler/ClassPatch/ThrowablePatch.php
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        return false;
+    }
+
+    /**
+     * @param string $type
+     * @return bool
+     */
+    private function isThrowable($type)
```
stof: Do we really need that many private methods ?

src/Prophecy/Doubler/ClassPatch/ThrowablePatch.php
```diff
@@ -0,0 +1,122 @@
...
+        $this->setParentClassToException($node);
+    }
+
+    /**
+     * @param ClassNode $node
+     */
+    private function checkItCanBeDoubled(ClassNode $node)
+    {
+        $className = $node->getParentClass();
+        if ($className != 'stdClass') {
```
stof: should use strict comparison

src/Prophecy/Doubler/ClassPatch/ThrowablePatch.php
```diff
@@ -0,0 +1,122 @@
...
+     * @return void
+     */
+    public function apply(ClassNode $node)
+    {
+        $this->checkItCanBeDoubled($node);
+        $this->setParentClassToException($node);
+    }
+
+    /**
+     * @param ClassNode $node
```
stof: useless phpdoc. It duplicates the typehint

ciaranmcnulty: Wasn't sure what the project policy was :-) will strip a few out in other places too then

src/Prophecy/Doubler/ClassPatch/ThrowablePatch.php
ciaranmcnulty: I tend to write this way but if you don't like it I can combine a few of them? Maybe it'll look better without the unnecessary docblocks

stof:
Tests are failing on PHP 5.x. The specs need to be skipped on 5.x when they expect the Throwable interface to exist

fixtures/ThrowableInterface.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace Fixtures\Prophecy;
+
+interface ThrowableInterface extends \Throwable
+{
+}
```
dkarlovi: Newline missing.

spec/Prophecy/Doubler/ClassPatch/ThrowablePatchSpec.php
```diff
@@ -0,0 +1,98 @@
...
+    {
+        if (\PHP_VERSION_ID < 70000) {
+            throw new SkippingException('Throwable is not defined in PHP 5');
+        }
+
+        $node->getParentClass()->willReturn('ArrayObject');
+
+        $this->shouldThrow('Prophecy\Exception\Doubler\ClassCreatorException')->duringApply($node);
+    }
+}
```
dkarlovi: Newline missing.

dkarlovi:
Hopefully, this allows https://github.com/symfony/symfony/pull/26702 to be re-applied. /cc @dunglas @nicolas-grekas

ciaranmcnulty:
@stof can you take another look please?

PR #417, opened 2018-10-17
Introduce Argument::withAllEntries() and Argument::containingAllOf() helpers

filakhtov:
Quite often seeing a need to check that array argument has a number of values or (key, value) pairs. Decided to put together two helper methods on the `Argument` class to help with these tasks:

```php
Argument::withAllEntries(['key' => 'value', 'foo' => 'bar']);
```
will match an array that has two specified (key, value) pairs: (`key`, `value`) and (`foo`, `bar`).

```php
Argument::containingAllOf('a', 'b', 'c');
```
will match an array that has all of the values specified as arguments.

### DIFF
--- a/spec/Prophecy/ArgumentSpec.php
+++ b/spec/Prophecy/ArgumentSpec.php
@@ -104,4 +104,16 @@ class ArgumentSpec extends ObjectBehavior
         $token = $this->approximate(10);
         $token->shouldBeAnInstanceOf('Prophecy\Argument\Token\ApproximateValueToken');
     }
+
+    function it_has_a_shortcut_for_array_entries_have_all_values()
+    {
+        $token = $this->withAllEntries(array('key1' => 'value1', 'key2' => 'value2'));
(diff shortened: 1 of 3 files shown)


ciaranmcnulty:
Not sure about the naming, they don't even seem too consistent with each other - have you any other suggestions for names?

filakhtov:
@ciaranmcnulty perhaps we can do something like `withAllPairs()` for key-value one and `withAllValues()` for checking values. Thoughts?

ciaranmcnulty:
Sorry this conversation is going so slowly - Prophecy is in need of a bit more TLC and I'm hoping I can spend more time on it.

Not sure Pairs is correct?

filakhtov:
By pair I mean key-value pair in this particular case. Do you have any better suggestions? We can do originally suggested `willAllEntries()` for this one and just rename `containingAllOf()` to `withAllValues()`.

ciaranmcnulty:
OMG this has been neglected (by us) for so long - sorry about that. The later naming sounds good to me if you can update/rebase?

ciaranmcnulty:
Closing for now, feel free to re-open

PR #419, opened 2018-11-02
Bumped php minimum version to 5.6

DonCallisto:
We should also change repo description (it states PHP5.3+)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,16 +14,12 @@ branches:
 
 matrix:
   include:
-    - php: 5.4
-    - php: 5.5
     - php: 5.6
       env: DEPENDENCIES='low'
     - php: 5.6
     - php: 7.0
     - php: 7.1
     - php: 7.2
-    - php: 5.3
-      dist: precise
   fast_finish: true
 
 install:
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ class UserTest extends PHPUnit_Framework_TestCase
 
 ### Prerequisites
 
-Prophecy requires PHP 5.3.3 or greater.
+Prophecy requires PHP 5.6 or greater.
 
 ### Setup through composer
 
@@ -187,14 +187,9 @@ you'll use promises for that:
 ```php
 $user->getName()->willReturn(null);
 
-// For PHP 5.4
+// For PHP 5.6
 $user->setName('everzet')->will(function () {
     $this->getName()->willReturn('everzet');
-});
-
-// For PHP 5.3
-$user->setName('everzet')->will(function ($args, $user) {
-    $user->getName()->willReturn('everzet');
 });
 
 // Or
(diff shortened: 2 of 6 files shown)


ciaranmcnulty:
Was there some related discussion around this, or is it a proposal from you?

DonCallisto:
It’s a proposal as PHP5.3 is no longer supported.

Even PHP5.6 and 7 will soon reach end of support.

http://php.net/supported-versions.php

DonCallisto:
Any updates for this? Maybe it will be safe to bump to 7.1 directly.
WDYT?

fezfez:
Any news on this ? Maybe bump to 7.2 directly now...

DonCallisto:
Yes, I think it would be better to raise it to 7.2 but no one seems to care anymore about prophecy.

fezfez:
... i think the maintainers are just busy

DonCallisto:
Don't get me wrong, I'm not here to complain anybody, I'm perfectly aware of OSS mechanisms.
Just saying my impression is that, as Prophecy is pretty stable, no one still spend time on it.
Look at this PR: one year has passed and no discussion nor review (or reject) has been done.
It is a pretty small PR, don't you think?
Take a look also at this https://github.com/phpspec/prophecy/pull/420, I'm willing to cover the PHPUnit tests in order to accomplish what Stof required but no feedback were given.
That's totally OK from my standpoint but my feeling is that no one cares anymore, totally subjective I know, but I think I'm authorized to think so without accuse anyone.

ciaranmcnulty:
Sorry for lack of review, I don't see a compelling reason to bump this for now. The amount of technical debt removed doesn't really warrant it.

If there's something we can't support easily, or if we are making other BC breaks compelling a new major I'd support a bump

DonCallisto:
@ciaranmcnulty don't worry.

If I can express my opinion, we should definitely bump the minimum version of PHP.

Main reasons:

- Discouraging who depends on prophecy to adopt older PHP version (I see it as a "compelling" reason to be honest; PHP 5.3 is even not listed [supported version](https://www.php.net/supported-versions.php) anymore)

- Declaration of intent: bumping PHP version and releasing along (if needed, otherwise keep it on master for the moment) a minor version is a declaration of intent: for the next (major) versions we'll not support older version anymore, be prepared.

Ecosystem evolves also for those reasons, I think we should keep pushing it.
Moreover stop to think about CI tests: an unaware contributor could break something that needs a quite amount of work in order to be OK with all PHP versions supported, he opens a PR and no one is able to point him to this discussion or, worst, he just doesn't understand if it's the case to bump or not the min. version because is

> something we can't support easily

I mean, dependencies are declarations of intent aswell, if it were the first time I see this project and it still supports PHP5.3 I would think that is a conscious choice, not a "non compelling" reason or that

> The amount of technical debt removed doesn't really warrant it.

Those are my questionable two cents.
WDYT?

stof:
>     * Discouraging who depends on prophecy to adopt older PHP version (I see it as a "compelling" reason to be honest; PHP 5.3 is even not listed [supported version](https://www.php.net/supported-versions.php) anymore)

being a test tool, we won't have any influence on which version project support as their min version. What will happen is that they will stop updating Prophecy if they still need to run their testsuite on old versions, potentially asking us to maintain them

DonCallisto:
On the other side, if any bug raises up in an older version, would not we still in "charge" to fix it?
I'm still not 100% convinced this is the right move as 

> stop updating Prophecy if they still need to run their testsuite on old versions, potentially asking us to maintain them

would be resolved by saying that older release are not longer supported and to update to fresh versions.

I'm not that experienced in OSS maintenance so I trust you 100% if the question I've raised with this comment doesn't bother you.

Issue #96, opened 2014-03-27
High memory usage in phpspec because of repeated double definitions


acoulton:
Hi,

I'm using phpspec and prophecy on a few projects, and the suite for one is now growing to a reasonable size with the result that today we started to hit out of memory errors during the suite run.

Our suite overall has 1205 examples and makes a lot of use of prophecy throughout, and I'm getting runtime of about 37 seconds and memory usage of about 178Mb.

I did some profiling and found that the biggest component is that prophecy generates a new double class definition for every mock - thanks to [the doubler](https://github.com/phpspec/prophecy/blob/master/src/Prophecy/Doubler/Doubler.php#L120) and the [unique sequence number generated for each double](https://github.com/phpspec/prophecy/blob/master/src/Prophecy/Doubler/Doubler.php#L120).

I don't think this should be necessary - surely the mocked class will always have the same methods and interfaces for duration of one suite run, so the double definition could just be built once?

I quickly experimented with removing the sequence counter and updating the doubler to do this:

``` php
        $name = $this->namer->name($class, $interfaces);
        if (class_exists($name, FALSE)) {
        return $name;
        }
        // otherwise continue as before to define and create the double
```

Suite runtime was reduced to 22 seconds (just over 50%) and memory usage to 50Mb (a quarter). 

The only issue I hit was this change also means that some == comparators in ExactValueToken, ObjectStateToken and ObjectProphecy encounter a recursive property and fail with a fatal during the comparison. Switching them to === works, but is obviously not appropriate for the matchers at least. 

It looks like this is because previously the doubles nested within those method prophecies and argument wildcards had different classes so == never had to recurse into comparing their properties, and now that it does it gets caught on the circular references between double and doubler. I can investigate further and look for a BC fix.

Before I spend more time on this, though, I wanted to check with those who know more:
- is there a reason for generating the double class each time that I'm not aware of?
- would you accept a fix along these lines into the project given appropriate tests and backwards compatability?

I saw that there is also a (currently unused) CachedDoubler implementation - which has the same issues with recursive property comparison - but this obviously then means managing an instance of that everywhere and passing it into Prophet instances. Since the existence or not of a class is part of global state anyway, it feels more logical to me just to generate the same name for each type of double and use class_exists to manage it.

DonCallisto:
Problem with `CachedDoubler` and `==` is that `Double` has `ObjectProphecy` in itself and `ObjectProphecy` has the double inside so we have a circular reference.
[...]

PR #420, opened 2018-11-03
Introduced and fixed CachedDoubler behaviour in order to reduce memor…

DonCallisto:
…y usage

With this PR we resolve https://github.com/phpspec/prophecy/issues/96 and https://github.com/phpspec/phpspec/issues/884.

Our empirical tests:

~ 6000 PHPSpec examples

Before patch:
- 1GB 900MB memory consumption
- 31676 declared classes
- 105 seconds per suite execution

After patch:
- 146MB (-1.301 %)
- 3148 declared classes (-1.006 %)
- 9 seconds per suite execution (- 1.166 %)

Only thing that should be considered here is the static-ness nature of `CachedDoubler`. 
Is this fine or is it improvable?

### DIFF
--- a/src/Prophecy/Doubler/CachedDoubler.php
+++ b/src/Prophecy/Doubler/CachedDoubler.php
@@ -51,8 +51,16 @@ class CachedDoubler extends Doubler
         foreach ($interfaces as $interface) {
             $parts[] = $interface->getName();
         }
+        foreach ($this->getClassPatches() as $patch) {
+            $parts[] = get_class($patch);
+        }
         sort($parts);
 
         return md5(implode('', $parts));
+    }
+
+    public function resetCache()
+    {
+        self::$classes = array();
     }
 }
--- a/src/Prophecy/Doubler/CachedDoubler.php
+++ b/src/Prophecy/Doubler/CachedDoubler.php
@@ -21,17 +21,7 @@ use ReflectionClass;
  */
 class CachedDoubler extends Doubler
 {
-    private $classes = array();
-
-    /**
-     * {@inheritdoc}
-     */
-    public function registerClassPatch(ClassPatch\ClassPatchInterface $patch)
-    {
-        $this->classes[] = array();
-
-        parent::registerClassPatch($patch);
-    }
+    private static $classes = array();
 
     /**
      * {@inheritdoc}
@@ -39,11 +29,11 @@ class CachedDoubler extends Doubler
     protected function createDoubleClass(ReflectionClass $class = null, array $interfaces)
     {
         $classId = $this->generateClassId($class, $interfaces);
-        if (isset($this->classes[$classId])) {
-            return $this->classes[$classId];
+        if (isset(self::$classes[$classId])) {
+            return self::$classes[$classId];
         }
 
-        return $this->classes[$classId] = parent::createDoubleClass($class, $interfaces);
+        return self::$classes[$classId] = parent::createDoubleClass($class, $interfaces);
     }
 
     /**
--- a/spec/Prophecy/Argument/ArgumentsWildcardSpec.php
+++ b/spec/Prophecy/Argument/ArgumentsWildcardSpec.php
@@ -14,7 +14,10 @@ class ArgumentsWildcardSpec extends ObjectBehavior
         $class = get_class($object->getWrappedObject());
         $hash  = spl_object_hash($object->getWrappedObject());
 
-        $this->__toString()->shouldReturn("exact(42), exact(\"zet\"), exact($class:$hash Object (\n    'objectProphecy' => Prophecy\Prophecy\ObjectProphecy Object (*Prophecy*)\n))");
+        $objHash = "exact(42), exact(\"zet\"), exact($class:$hash Object (\n    'objectProphecyClosure' => Closure:%s Object (\n        0 => Closure:%s Object\n    )\n))";
+
+        $hashRegexExpr = '[a-f0-9]{32}';
+        $this->__toString()->shouldMatch(sprintf('/^%s$/', sprintf(preg_quote("$objHash"), $hashRegexExpr, $hashRegexExpr)));
     }
 
     function it_generates_string_representation_from_all_tokens_imploded(
--- a/spec/Prophecy/Argument/Token/ExactValueTokenSpec.php
+++ b/spec/Prophecy/Argument/Token/ExactValueTokenSpec.php
@@ -125,13 +125,15 @@ class ExactValueTokenSpec extends ObjectBehavior
 
     function it_generates_proper_string_representation_for_object(\stdClass $object)
     {
-        $objHash = sprintf('%s:%s',
+        $objHash = sprintf('exact(%s:%s',
             get_class($object->getWrappedObject()),
             spl_object_hash($object->getWrappedObject())
-        );
+        ) . " Object (\n    'objectProphecyClosure' => Closure:%s Object (\n        0 => Closure:%s Object\n    )\n))";
 
         $this->beConstructedWith($object);
-        $this->__toString()->shouldReturn("exact($objHash Object (\n    'objectProphecy' => Prophecy\Prophecy\ObjectProphecy Object (*Prophecy*)\n))");
+
+        $hashRegexExpr = '[a-f0-9]{32}';
+        $this->__toString()->shouldMatch(sprintf('/^%s$/', sprintf(preg_quote("$objHash"), $hashRegexExpr, $hashRegexExpr)));
     }
 }
 
--- a/spec/Prophecy/Argument/Token/IdenticalValueTokenSpec.php
+++ b/spec/Prophecy/Argument/Token/IdenticalValueTokenSpec.php
@@ -141,12 +141,14 @@ class IdenticalValueTokenSpec extends ObjectBehavior
 
     function it_generates_proper_string_representation_for_object($object)
     {
-        $objHash = sprintf('%s:%s',
+        $objHash = sprintf('identical(%s:%s',
             get_class($object->getWrappedObject()),
             spl_object_hash($object->getWrappedObject())
-        );
+        ) . " Object (\n    'objectProphecyClosure' => Closure:%s Object (\n        0 => Closure:%s Object\n    )\n))";
 
         $this->beConstructedWith($object);
-        $this->__toString()->shouldReturn("identical($objHash Object (\n    'objectProphecy' => Prophecy\Prophecy\ObjectProphecy Object (*Prophecy*)\n))");
+
+        $hashRegexExpr = '[a-f0-9]{32}';
+        $this->__toString()->shouldMatch(sprintf('/^%s$/', sprintf(preg_quote("$objHash"), $hashRegexExpr, $hashRegexExpr)));
     }
 }
--- a/spec/Prophecy/Doubler/CachedDoublerSpec.php
+++ b/spec/Prophecy/Doubler/CachedDoublerSpec.php
@@ -1,0 +1,414 @@
+<?php
+
+namespace spec\Prophecy\Doubler;
+
+use PhpSpec\ObjectBehavior;
+use Prophecy\Doubler\ClassPatch\ClassPatchInterface;
+use Prophecy\Doubler\Generator\ClassCreator;
+use Prophecy\Doubler\Generator\ClassMirror;
+use Prophecy\Doubler\Generator\Node\ClassNode;
+use Prophecy\Doubler\NameGenerator;
+
+class CachedDoublerSpec extends ObjectBehavior
+{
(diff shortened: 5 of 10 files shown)


DonCallisto:
🎉 test are now green. 
WDYT?

DonCallisto:
@everzet @stof @ciaranmcnulty have you had the change to check this?
This is not only a performace improvement but also resolve a bug where we check objects with `==` as, at the moment, two different classes will always return `false` (as the double will be different) and so anyone would be able to write a use case for this in test.

src/Prophecy/Doubler/CachedDoubler.php
```diff
@@ -21,29 +21,19 @@
...
-    /**
-     * {@inheritdoc}
-     */
-    public function registerClassPatch(ClassPatch\ClassPatchInterface $patch)
-    {
-        $this->classes[] = array();
-
-        parent::registerClassPatch($patch);
-    }
+    private static $classes = array();
```
stof: should it really be static ? Registering different class patches in a different instance must not reusable the old class definition (and that's why `registerClassPatch` was resetting it)

DonCallisto:
> This would need some functional tests proving that 2 different doubles don't get considered as equal when matching arguments and so on

@stof instances are different whereas class definitios are the same. I've tested this patch against our 6000 test suite and all runs perfectly. I can - for sure - try those kind of tests when I have free time in these upcoming weeks.

> should it really be static ? Registering different class patches in a different instance must not reusable the old class definition (and that's why registerClassPatch was resetting it)

That's a doubt I had. So you can confirm that, in some situation, same class could be "patched" in a different manner?
One solution could be to include applyed patches (don't know if this breaks incapsulation, I need to take a depper look) in class hash.
Other solution is to try avoid the staticness nature of the `CachedDoubler`.

WDYT?

stof:
@DonCallisto that fact that we allow registering patchers in the Doubler instance means that 2 instances might have different patchers.
Including a hash of the registered patchers might indeed be a solution.

DonCallisto:
@stof just some basic considerations here and there before work on this solution.

To keep a simple design and in order to prevent iteraring two times on `patches`, creating two nodes etc, I would like to return frorm `createDoubleClass` something called `DoubleClassCreatedDTO(string $fqcn, array $appliedPatchesFQCNs)` that I can reuse in the `CachedDoubler`. This way the `Doubler` does not require some tricky and dirty hacks (like keep in memory something that helps to retrieve applied patches; this seems more a job for the `CachedDoubler`, isn't it?). 
This is somehow necessary, at least from my POV, in order to avoid re-creating the node two times in the same process and keep all actors well separated.
I haven't any other valid idea at the moment, do you?

One final note: how would you test this in a functional fashion? I've broke my head thinking about this but I'm currently to a dead end.
IIUC you would test, basically, `ExactValueToken` and `IdenticalToken` in a functional way (other arguments seems not to be "impacted" from this patch).
What I miss here is how can I actually test them.

If you give me a feedback on the first consideration and a pointer for the latter, I will be happy to work on this.

DonCallisto:
Moreover, thinking about a functional implementation, I’m still not sure we need that kind of test.

I mean, this patch will change only the way class *definitions* are created, not the instances.
As arguments works on instances, I’m pretty sure those tests have no value at all.

Am I missing something?

DonCallisto:
Maybe another option is to create a `CachedClassMirror` based on [`ClassMirror`](https://github.com/phpspec/prophecy/blob/aba3a3ac211ee21e5eb20781a296c859c4163ce6/src/Prophecy/Doubler/Generator/ClassMirror.php) and construct `CachedDoubler` with this particular type. 
In this way it could be safe to "demand" for "supported patches" from `CachedDoubler` without re-creating the `node` two times.
Only "issue" I see here is that this would require an override of `createDoubleClass` in `CachedDoubler` or to return a "cloned" `Node` (but that is what I'm trying to avoid) as `patches` in `Doubler` will be re-applied.
In both case I don't see any advantage and I would stay with my first idea.

WDYT?

stof:
@DonCallisto what is important is not the list of applied patches (which could not be determined to build the cache key, as you would still have to create the class and then discard it), but the list of registered patchers.
It is possible that 2 instances with different patchers would generate the same code for a double (because the extra patcher was not doing anything for that class), but I think that's fine. It is better to have cache misses sometimes for such case (which is not the common case anyway, as most projects use the same patchers for their whole testsuite) rather than having wrong cache hits (as that's a bug).

DonCallisto:
@stof whereas the idea behind applied/registered patch is not clear to me

> as you would still have to create the class and then discard it

(that's not true as we create the class *after* it is patched)

and

> It is possible that 2 instances with different patchers would generate the same code for a double (because the extra patcher was not doing anything for that class)

here there's still not instances, only definitions

I think that's fine to build with registered patches anyway because same class, with same interface and same registered patchers in the doubler will produce the same code. That means that there's no way for the same class to have applied one patch in a situation and not to be applied in other (unless the patch isn't registered, of course).

I've implemented this fix in a separate commit. I'm still writing some spec tests (as stated before, I don't think we need functional tests for this as we producing class definitions, not instances) to cover all the possibilities (permutation of same/different class, same/different/no interfaces, same/different/no patches).

Do it LGTY?

stof:
> (that's not true as we create the class after it is patched)

Having to call the parent logic to create the class was what I was talking about.
The updated versions of the PR is precisely doing what I was suggesting: talking all registered patchers into account for the hash.

> > It is possible that 2 instances with different patchers would generate the same code for a double (because the extra patcher was not doing anything for that class)
> here there's still not instances, only definitions

this sentence is talking about instances of the doubler, not instance of the double.

> I've implemented this fix in a separate commit. I'm still writing some spec tests (as stated before, I don't think we need functional tests for this as we producing class definitions, not instances) to cover all the possibilities (permutation of same/different class, same/different/no interfaces, same/different/no patches).

Last time I tried enabling the CachedDoubler in my tests, the argument matching (when using a double as argument in a method of another double) broke for some cases, because 2 doubles of the same class were now considered the same for argument matching (without the cached doubler, they were instances of different classes, which is why it worked).
My concern here is not that this PR produces a bad double class. My concern is that it breaks compatibility with some other parts of Prophecy.

DonCallisto:
@stof 

> because 2 doubles of the same class were now considered the same for argument matching

Yes, I understand your point but I don't think that the following statement is true

> because 2 doubles of the same class were now considered the same for argument matching

because instances are different so from my POV.

Yes, it could break some tests because `==` was never evaluated to `true` but that was a bug and will prevent users to write tests for such use case.

Could you try it against the code you broken when tried out `CachedDoubler`?
Moreover I don't think that internally we break anything from Prophecy POV. If you perceive we did, please point me out parts of Prophecy code so I can think more deeply about this possible issue and eventually try to find a solution.

Thanks in advance for your review and your time.

DonCallisto:
Any update? I would be happy to try and see what Stof claims but I really can't understand how to make (and on what) the functional test :\

DonCallisto:
Any news? I'm willing to continue my work but without any clue I just can't write litterally any line of those tests that stof mentioned above

DonCallisto:
@stof maybe the tests you’re waiting for are something like [this](url
https://github.com/bsod85/prophecy/blob/memory_optimization/tests/ProphetTest.php) repeated for all interesting matchers on doubles?

All the credits goes to my colleague [@bsod85](url
https://github.com/bsod85)

Please let us know if this is acceptable (there is something that need to be fixed, we know, but the idea behind it should be clear) or if is not what you expect to be tested.

Thanks!

DonCallisto:
@stof Sorry to disturb you, could you give me any update on this. I know that probably you are busy but I would like only to have a feedback, even if this is "I can't take a look at this for a while".
Moreover I would like other contributors like @everzet or @ciaranmcnulty to take a look at this and the whole conversation and give any opinion if possibile.

Thank you in advance guys.

romm:
Hi @DonCallisto I just tested your patch and could see a great performance improvement indeed, for a suite of 221 tests with 591 assertions.

Without the patch:
> Time: 3.59 seconds, Memory: 32.00MB

With the patch on:
> Time: 1.17 seconds, Memory: 22.00MB

However, a test of mine is broken with the patch applied:

```php
/** @test */
public function fetch_service_returns_service_with_highest_priority()
{
    /* *** Initialization *** */
    $source = [];
    $targetType = $this->prophesize(Type::class)->reveal();

    $serviceA = $this->prophesize(Service::class);
    $serviceA->priority()
        ->shouldBeCalled()
        ->willReturn(100);

    $serviceB = $this->prophesize(Service::class);
    $serviceB->priority()
        ->shouldBeCalled()
        ->willReturn(200);

    /* *** Process *** */
    $this->registry->register(get_class($serviceA->reveal()));
    $this->registry->register(get_class($serviceB->reveal()));

    $service = $this->registry->fetchFor($source, $targetType);

    /* *** Assertion *** */
    $this->assertInstanceOf(get_class($serviceB->reveal()), $service);
}
```

Maybe I shouldn't have done the assertion based on the class of the double, but as far as I can see your patch doesn't allow testing cases like this one, maybe am I missing something?

DonCallisto:
@romm What does `get_class` returns in this example? Please provide a var dump that helps me to understand better.
Thanks 🙂

romm:
@DonCallisto adding this dumps:

```php
var_dump(get_class($serviceA->reveal()));
var_dump(get_class($serviceB->reveal()));
```

I get the following values:

Without your patch
> string(17) "Double\Service\P4"
> string(17) "Double\Service\P5"

With your patch
> string(17) "Double\Service\P2"
> string(17) "Double\Service\P2"

After some more analysis, the services are instantiated with the given class and that's probably not the best way to do it: an instance of the object should be given instead.

So I think I'm going to change the implementation anyway, but you better know what was going on here.

DonCallisto:
@romm  Yes, it is was I suspected without seeing you implementation. That's a case where Prophecy didn't catch a possible problem with your implementation due to different class names. This patch will correct this scenario also.

DonCallisto:
@ciaranmcnulty could you please take a look at this? We're basically stuck but I feel like this PR could be merged with right PHPUnit tests done (see https://github.com/phpspec/prophecy/pull/420#issuecomment-451598411)

DonCallisto:
@everzet could you take a look at this? It's four months since this PR is open and I feel like we're there or at least very close to mergeable code.

src/Prophecy/Doubler/CachedDoubler.php
Ocramius: I'd say that the PHP engine already has a set of registered classes: if we can do a lookup via hash name (`class_exists($hash, false)`) that should be enough

DonCallisto: @Ocramius I can modify this but this PR seems to be "sleepy" and no one seems to care to get rid of that memory leak.

Ocramius: Attention has been drawn here from https://github.com/Roave/no-leaks/pull/14 ;-)

DonCallisto: Yes, I read the thread. 

I mean, if neither @stof nor @ciaranmcnulty provide their feeback, following with a possible merge, this PR will stuck here forever. 

I know that everybody is busy and this is OSS, but having those kind of feedback (like yours that is only an enhancement because I feel it doesn't really change the sense of this PR) is crucial to move to merging direction.

I can try in the following hours to make your fix but @stof was worried about other kind of issues (read below) not related to staticness of class hashes.

romm:
Hello, any news concerning this PR? Could we help to get it merged somehow? :)

DonCallisto:
@romm it seems not

DonCallisto:
@ciaranmcnulty as I see you're quite active these days, could you try to take a look a this and thread?
It would be great.
Thanks.

ciaranmcnulty:
@stof do you have some time to continue the conversation here? I am not clear what the issues you observed were

DonCallisto:
Just to provide more context, an interesting question never answered was [this](https://github.com/phpspec/prophecy/pull/420#issuecomment-451598411).
From what I understood, @stof was worried about such things.
Don't know if what we're testing there is "the right thing" to clear any doubt and, if so, if those kind of tests should be included in this PR.
Still willing to work on this if we clear out what points needed to be tested and what kind of tests should I provide.

PS.: I'm using since a year a fork of mine with modification I've provided in this PR. Old examples (as pointed out in first comment) ~ 6000 didn't broke and new examples (can't quantify them) worked as expected.

ciaranmcnulty:
OK I'm going to merge this, we can always back it out if users find issues

We aren't maintaining actively enough to get into long 'what if' conversations if they're not resolved

DonCallisto:
Thanks @ciaranmcnulty!
Could you please release a tagged version with this merge?

PR #422, opened 2018-11-09
Bump phpdocumentor/reflection-docblock to ^5.0

### LINKED ISSUE #415: Add ^5 to allowed versions of phpdocumentor/reflection-docblock
Version 5 of reflection-docblock is coming out soon, and currently code that uses its new features will not be compatible with prophecy.

ricpelo:
Fixes #415.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
 
     "require": {
         "php":                               "^5.3|^7.0",
-        "phpdocumentor/reflection-docblock": "^2.0|^3.0.2|^4.0",
+        "phpdocumentor/reflection-docblock": "^2.0|^3.0.2|^4.0|^5.0",
         "sebastian/comparator":              "^1.1|^2.0|^3.0",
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
As the tests pass, this seems harmless right? Thanks @ricpelo

Miclef1:
> As the tests pass, this seems harmless right? Thanks @ricpelo

This is not harmless.
You have introduced a new major version dependency with a minor version update of your project, which - to my understanding - violates the principle of semantic versioning.

We currently have an issue with this.
See following error message:
`phpdocumentor/reflection-docblock 5.0.0 requires ext-filter ^7.1 -> the requested PHP extension filter has the wrong version (7.2.12) installed.`

stof:
We allow the new version of the dependency. But we don't force using it. Composer will still resolve compatible dependencies.

stof:
btw, the message looks weird. `7.2.12` matches the `^7.1` constraint.

PR #425, opened 2018-11-23
[README] explain how to return different values for the same call

lyrixx:


README.md
```diff
@@ -165,6 +165,13 @@ promise, there's plenty others you can use:
 Keep in mind, that you can always add even more promises by implementing
 `Prophecy\Promise\PromiseInterface`.
 
+If the method is called sereval times, you can use the following syntaxe to return different
```
odolbeau: syntax

README.md
```diff
@@ -165,6 +165,13 @@ promise, there's plenty others you can use:
 Keep in mind, that you can always add even more promises by implementing
 `Prophecy\Promise\PromiseInterface`.
 
+If the method is called sereval times, you can use the following syntax to return different
```
Taluu:
```suggestion
If the method is called several times, you can use the following syntax to return different
```

README.md
```diff
@@ -165,6 +165,13 @@ promise, there's plenty others you can use:
...
+values for each calls:
```
Pierstoval:
```suggestion
values for each call:
```

stof:
I would rather document it as a note at the end of the `Method prophecies idempotency` section, as this feature is actually not recommended for most cases (relying on the order of calls for the same arguments tends to make test fragile, as adding one more call can break everything)

lyrixx:
@stof You are right. I updated the PR

gquemener:
Allow me to share this conversation where Konstantin gives more details about why this is, most of the time, not desirable to use such feature: https://gist.github.com/gquemener/292e7c5a4bbb72fd48a8

lyrixx:
I often see people asking for that.  Could we merge this one since there is a warning?
Thanks

lyrixx:
I often see people asking for that.  Could we merge this one since there is a warning?
Thanks

lyrixx:
ping

README.md
```diff
@@ -206,6 +206,17 @@ $user->setName('everzet')->will(function ($args) use ($user) {
...
+Note: If the method is called several times, you can use the following syntax to return different
+values for each call:
+
+```php
+$prophecy->read('123')->willReturn(1, 2, 3);
+```
+
+This feature is actually not recommended for most cases (relying on the order of
+calls for the same arguments tends to make test fragile, as adding one more call
+can break everything).
```
jakzal: I think the sentence in parenthesis should  just be a regular sentence. It's an important note.

lyrixx: @jakzal Thanks for your feedback. I removed parenthesis

PR #429, opened 2019-03-19
Allow `MethodProphecy::willThrow()` to accept Throwable as string

### LINKED ISSUE #428: `MethodProphecy::willThrow()` doesn't accept (an extension of) `Throwable`
Working on implementing unit tests for a piece of code that uses a PSR-18 client I was unable to have a `\Psr\Http\Client\ClientInterface::sendRequest()` MethodProphecy to throw one of the expected exception interface. To reproduce:

```php
<?php

class ReproductionTest extends \PHPUnit\Framework\TestCase
{
    public function testThrowable()
    {
        $client = $this->prophesize(\Psr\Http\Client\ClientInterface::class);
        $client->sendRequest(Argument::type(\Psr\Http\Message\RequestInterface::class))
            ->willThrow(\Psr\Http\Client\ClientExceptionInterface::class);
        
        // do assertions
    }
}
```

Since `\Psr\Http\Client\ClientExceptionInterface` extends `\Throwable` this _should_ work. But it doesn't:

```bash
Prophecy\Exception\InvalidArgumentException : Exception / Throwable class or instance expected as argument to ThrowPromise, but got Psr\Http\Client\ClientExceptionInterface.
```

This scenario is reproducible on the latest stable version `1.8.0` and on the `master` branch.

timoschinkel:
- add additional check for existence of interface in contructor
- change `is_subclass_of` to `is_a` to allow for `Throwable` and subclasses of `Throwable`
- add specs for cases of `Throwable` and extensions of `Throwable`

To maintain support for php 5.3 the new specs are only applicable for PHP 7.0 and higher and `ThrowPromise::isValidThrowable()` checks for both `Exception` and `Throwable`.

fixes #428

### DIFF
--- a/spec/Prophecy/Promise/ThrowPromiseSpec.php
+++ b/spec/Prophecy/Promise/ThrowPromiseSpec.php
@@ -84,6 +84,28 @@ class ThrowPromiseSpec extends ObjectBehavior
 
         $this->shouldNotThrow('Prophecy\Exception\InvalidArgumentException')->duringInstantiation();
     }
+
+    function it_throws_an_extension_of_throwable_by_class_name()
+    {
+        if (!interface_exists('\Throwable')) {
+            throw new SkippingException('The interface Throwable, introduced in PHP 7, does not exist');
+        }
+
+        $this->beConstructedWith('\Fixtures\Prophecy\ThrowableInterface');
+
+        $this->shouldNotThrow('Prophecy\Exception\InvalidArgumentException')->duringInstantiation();
+    }
+
+    function it_throws_a_throwable_by_class_name()
+    {
+        if (!interface_exists('\Throwable')) {
+            throw new SkippingException('The interface Throwable, introduced in PHP 7, does not exist');
+        }
+
+        $this->beConstructedWith('\Throwable');
+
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
Thanks for the fix!

spec/Prophecy/Promise/ThrowPromiseSpec.php
```diff
@@ -84,6 +84,28 @@ function it_throws_an_exception_by_class_name()
 
         $this->shouldNotThrow('Prophecy\Exception\InvalidArgumentException')->duringInstantiation();
     }
+
+    function it_throws_an_extension_of_throwable_by_class_name()
```
drupol: Typo: s/extension/exception

timoschinkel: It's been a while since I filed this PR - and it's merged already - but I think I do mean "extension" here. Before you could do `willThrow(\Throwable::class)`, but not with interfaces that extend `\Throwable`. Hence the usage of `\Fixtures\Prophecy\ThrowableInterface`; that fixture extends `\Throwable`. The explanation is in the linked issue.

drupol: Fair enough !

PR #430, opened 2019-03-23
Proposal/qa improvements

belgattitude:
Just an initial idea:

- removed compat for < php 5.6
- increased to deps
- support for phpunit 8
- introduced phpstan level 1
- few minor fixes (autoload-dev)
- travis: reworked tu use stages
- travis: added roave/no-leaks (phpunit 8 / php 7.3)
- travis: added phpstan level 1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,12 @@ jobs:
       script:
         - composer require --dev phpstan/phpstan phpstan/phpstan-unit
         - composer phpstan
+    - stage: No-leaks
+      php: 7.3
+      env: XDEBUG=true
+      script:
+          - composer require --dev roave/no-leaks
+          - vendor/bin/roave-no-leaks
     - stage: Tests PHP 7.2
       php: 7.2
       env: XDEBUG=true
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,26 +12,83 @@ branches:
     - /^feature\/.*$/
     - /^optimization\/.*$/
 
-matrix:
-  include:
-    - php: 5.4
-    - php: 5.5
-    - php: 5.6
-      env: DEPENDENCIES='low'
-    - php: 5.6
-    - php: 7.0
-    - php: 7.1
-    - php: 7.2
-    - php: 7.3
-    - php: 5.3
-      dist: precise
-  fast_finish: true
+before_install:
+  - |
+    if [ "x$XDEBUG" != "xyes" ]; then
+      mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{,.disabled} || true
+    fi
 
 install:
-  - export COMPOSER_ROOT_VERSION=dev-master
-  - if [ "$DEPENDENCIES" != "low" ]; then composer update; fi;
-  - if [ "$DEPENDENCIES" == "low" ]; then composer update --prefer-lowest; fi;
+  - |
+    if [[ "x$DEPS" == "xlowest" ]]; then
+      travis_retry composer -n update --prefer-dist --prefer-lowest --prefer-stable --no-interaction $COMPOSER_ARGS;
+    else
+      travis_retry composer -n update --prefer-dist --no-interaction $COMPOSER_ARGS;
+    fi
 
-script:
-  - vendor/bin/phpspec run -fpretty -v
-  - vendor/bin/phpunit
+jobs:
+  allow_failures:
+    - php: nightly
+    - stage: No-leaks
+
+  include:
+    - stage: Tests PHP 7.3
+      php: 7.3
+      env: XDEBUG=yes
+      script:
+        - vendor/bin/phpspec run -fpretty -v
+        - vendor/bin/phpunit
+    - stage: PHPStan
+      php: 7.3
+      env: XDEBUG=false
+      script:
+        - composer require --dev phpstan/phpstan phpstan/phpstan-unit
+        - composer phpstan
+    - stage: Tests PHP 7.2
+      php: 7.2
+      env: XDEBUG=true
+      script:
+        - vendor/bin/phpspec run -fpretty -v
+        - vendor/bin/phpunit
+    - stage: Tests PHP 7.1
+      php: 7.1
+      env: XDEBUG=true
+      script:
+        - vendor/bin/phpspec run -fpretty -v
+        - vendor/bin/phpunit
+    - stage: Tests PHP 7.1 (lowest)
+      php: 7.1
+      env: XDEBUG=true DEPS=lowest
+      script:
+        - vendor/bin/phpspec run -fpretty -v
+        - vendor/bin/phpunit
+
+    # End of life versions
+
+    - stage: Tests PHP 7.0
+      php: 7.0
+      env: XDEBUG=true
+      script:
+        - vendor/bin/phpspec run -fpretty -v
+        - vendor/bin/phpunit
+    - stage: Tests PHP 5.6
+      php: 5.6
+      env: XDEBUG=true
+      script:
+        - vendor/bin/phpspec run -fpretty -v
+        - vendor/bin/phpunit
+    - stage: Tests PHP 5.6 (lowest deps)
+      php: 5.6
+      env: XDEBUG=true DEPS=lowest
+      script:
+        - vendor/bin/phpspec run -fpretty -v
+        - vendor/bin/phpunit
+
+    # Nightly builds
+
+    - stage: Nightlty
+      php: nightly
+      env: XDEBUG=false COMPOSER_ARGS=--ignore-platform-reqs
+      script:
+        - vendor/bin/phpspec run -fpretty -v
+        - vendor/bin/phpunit
--- a/composer.json
+++ b/composer.json
@@ -18,16 +18,16 @@
     ],
 
     "require": {
-        "php":                               "^5.3|^7.0",
-        "phpdocumentor/reflection-docblock": "^2.0|^3.0.2|^4.0",
-        "sebastian/comparator":              "^1.1|^2.0|^3.0",
+        "php":                               "^5.6|^7.0",
+        "phpdocumentor/reflection-docblock": "^3.0|^4.0",
+        "sebastian/comparator":              "^1.2|^2.0|^3.0",
         "doctrine/instantiator":             "^1.0.2",
-        "sebastian/recursion-context":       "^1.0|^2.0|^3.0"
+        "sebastian/recursion-context":       "^2.0|^3.0"
     },
 
     "require-dev": {
         "phpspec/phpspec": "^2.5|^3.2",
-        "phpunit/phpunit": "^4.8.35 || ^5.7 || ^6.5 || ^7.1"
+        "phpunit/phpunit": "^5.7 || ^6.5 || ^7.1 || ^8.0"
     },
 
     "autoload": {
@@ -38,8 +38,13 @@
 
     "autoload-dev": {
         "psr-4": {
+            "Tests\\Prophecy\\": "tests",
             "Fixtures\\Prophecy\\": "fixtures"
         }
+    },
+
+    "scripts": {
+        "phpstan": "vendor/bin/phpstan analyse -c phpstan.neon src tests"
     },
 
     "extra": {
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# EditorConfig is awesome: http://EditorConfig.org
+# top-most EditorConfig file
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+
+[*.neon]
+indent_style = tab
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 /composer.lock
 /vendor
 /phpunit.xml
+.phpunit.result.cache
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,0 +1,22 @@
+includes:
+	- vendor/phpstan/phpstan-phpunit/extension.neon
+	- vendor/phpstan/phpstan-phpunit/rules.neon
+parameters:
+	level: 1
+	fileExtensions:
+		- php
+	reportUnmatchedIgnoredErrors: true
+	ignoreErrors:
+		# For legacy compat
+		-
+			message: '#Return typehint of method Prophecy\\PhpDocumentor\\ClassAndInterfaceTagRetriever::getTagList\(\) has invalid type phpDocumentor\\Reflection\\DocBlock\\Tag\\MethodTag.#'
+			path: src/Prophecy/PhpDocumentor/ClassAndInterfaceTagRetriever.php
+		-
+			message: '#Return typehint of method Prophecy\\PhpDocumentor\\LegacyClassTagRetriever::getTagList\(\) has invalid type phpDocumentor\\Reflection\\DocBlock\\Tag\\MethodTag.#'
+			path: src/Prophecy/PhpDocumentor/LegacyClassTagRetriever.php
+		-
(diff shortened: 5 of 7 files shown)


belgattitude:
A little note about phpstan:

Till support for php < 7.1 is there, manual installation is required to use it locally.

```php
composer require phpstan/phpstan phpstan/phpstan-phpunit --dev
composer phpstan
```

Travis actually does this for all versions >= 7.1 automatically. But it's not ideal for developers, too easy to commit composer.json by mistake.

Is supporting php < 7.1 still makes sense today ?

.travis.yml
```diff
@@ -12,26 +12,93 @@ branches:
...
-    - php: 7.0
-    - php: 7.1
-    - php: 7.2
-    - php: 7.3
-    - php: 5.3
-      dist: precise
-  fast_finish: true
+before_install:
+  - |
+    if [ "x$XDEBUG" != "xyes" ]; then
```
Taluu: why care about xdebug ? Disable it, period.

.travis.yml
```diff
@@ -12,26 +12,93 @@ branches:
...
+        - vendor/bin/phpspec run -fpretty -v
+        - vendor/bin/phpunit
+    - stage: Tests PHP 7.1 (lowest)
+      php: 7.1
+      env: XDEBUG=false DEPS=lowest
+      script:
+        - vendor/bin/phpspec run -fpretty -v
+        - vendor/bin/phpunit
+
+    # End of life versions
```
Taluu: these should be removed imo. Even if it's allowed in the composer requirement : "it should be supported, but it's not guaranteed"

composer.json
```diff
@@ -18,16 +18,16 @@
...
+        "sebastian/comparator":              "^1.2|^2.0|^3.0",
         "doctrine/instantiator":             "^1.0.2",
-        "sebastian/recursion-context":       "^1.0|^2.0|^3.0"
+        "sebastian/recursion-context":       "^2.0|^3.0"
     },
 
     "require-dev": {
-        "phpspec/phpspec": "^2.5|^3.2",
-        "phpunit/phpunit": "^4.8.35 || ^5.7 || ^6.5 || ^7.1"
+        "phpspec/phpspec": "^3.0",
```
Taluu:
```suggestion
        "phpspec/phpspec": "^3.2",
```

?

composer.json
```diff
@@ -18,16 +18,16 @@
     ],
 
     "require": {
-        "php":                               "^5.3|^7.0",
-        "phpdocumentor/reflection-docblock": "^2.0|^3.0.2|^4.0",
-        "sebastian/comparator":              "^1.1|^2.0|^3.0",
+        "php":                               "^5.6|^7.0",
```
Taluu: We should remove EOL'd versions IMO.

.travis.yml
belgattitude: don't lock the future. coverage.  no leaks and possibly others. So it's already there. By default disabled, just saving few seconds, nothing to worry about.

.travis.yml
belgattitude: agree, will wait for other opinions. in case there reason to keep it.  anyway I'm feeling to refactor to php7.1 in a different p/r. the idea is to not put everything together.

composer.json
belgattitude: +1

composer.json
belgattitude: good catch thank ! I'll fix but it seems to work anyway.  stage with lowest reps.

belgattitude:
yes. stages are more verbose,  but makes adding different logics easier and isolated  phpstan, cs checks.... but you're right nothing justify it for now. I still think is the way to go.

belgattitude:
@Taluu 

> The travis file looks more complicated than what it should be IMO.

You're right what about the latest one ?

composer.json
```diff
@@ -38,10 +38,22 @@
...
     },
 
+    "scripts": {
+        "test": [
+            "@test:phpspec",
+            "@test:phpunit"
+        ],
+        "test:phpunit": "vendor/bin/phpunit",
+        "test:phpspec": "vendor/bin/phpspec run -fpretty -v",
+        "phpstan": "vendor/bin/phpstan analyse -c phpstan.neon src tests"
```
Taluu: IIRC, no need in composer to use the `vendor/...` path. You can use `phpunit` and it will be correctly resolved. And also, IMO, not really needed to declare those, you can directly call it...

(I'd rather use makefiles for that)

belgattitude: Nice catch, Thx it's fixed ;)

belgattitude:
Possible to merge ?

ciaranmcnulty:
@belgattitude It's been 2 days and dropping support for old versions is a big decision - give people a chance to catch up

CHANGES.md
```diff
@@ -1,3 +1,25 @@
+1.?.? / 2019/??/?? (NOT RELEASED)
+==================
+
+Minimum supported PHP version is now 5.6 *(5.3, 5.4 and 5.5 dropped)* and removed support 
+for `phpdocumentor/reflection-docblock: ^2.0` *(minimum is now ^3.1)*, 
+`sebastian/recursion-context: ^1.0` *(minimum is ^2.0)* and `sebastian/comparator: ^1.1` *(now ^1.2)*. 
+
+* [improved] func_get_args calls by variadic args: `Argument::allOf(...$tokenList)` [#430](https://github.com/phpspec/prophecy/pull/430).
```
stof: I don't think this needs to be documented in the changelog. It is purely internal change. The method had a variadic signture before, and it still has one now.

CHANGES.md
```diff
@@ -1,3 +1,25 @@
+1.?.? / 2019/??/?? (NOT RELEASED)
```
stof: Well, next 1.x version would be `1.9.0`. Dropping support for some PHP version is definitely not a good idea in a patch release. It deserves doing it in a minor version.

CHANGES.md
```diff
@@ -1,3 +1,25 @@
...
+for `phpdocumentor/reflection-docblock: ^2.0` *(minimum is now ^3.1)*, 
+`sebastian/recursion-context: ^1.0` *(minimum is ^2.0)* and `sebastian/comparator: ^1.1` *(now ^1.2)*. 
+
+* [improved] func_get_args calls by variadic args: `Argument::allOf(...$tokenList)` [#430](https://github.com/phpspec/prophecy/pull/430).
+* [improved] Removed obsolete eval() code for generators in `MethodProphecy` [#430](https://github.com/phpspec/prophecy/pull/430).
+* [added] Deps: support for phpunit 8.0 *(with few warnings till we drop support for PHP5)* [#430](https://github.com/phpspec/prophecy/pull/430).
+* [added] Dev: convenience composer script `composer test`, will run phpspec and phpunit [#430](https://github.com/phpspec/prophecy/pull/430). 
+* [added] Dev: phpstan level 4 + `composer phpstan` convenience script [#430](https://github.com/phpspec/prophecy/pull/430).
+* [fixed] Minor typehint fixes [#430](https://github.com/phpspec/prophecy/pull/430).
+* [fixed] Minor: autoload-dev includes `Tests' and 'spec' namespaces [#430](https://github.com/phpspec/prophecy/pull/430).
```
stof: you have a mix of backticks and single quotes here, which break the Yaml

CHANGES.md
```diff
@@ -1,3 +1,25 @@
...
+
+Minimum supported PHP version is now 5.6 *(5.3, 5.4 and 5.5 dropped)* and removed support 
+for `phpdocumentor/reflection-docblock: ^2.0` *(minimum is now ^3.1)*, 
+`sebastian/recursion-context: ^1.0` *(minimum is ^2.0)* and `sebastian/comparator: ^1.1` *(now ^1.2)*. 
+
+* [improved] func_get_args calls by variadic args: `Argument::allOf(...$tokenList)` [#430](https://github.com/phpspec/prophecy/pull/430).
+* [improved] Removed obsolete eval() code for generators in `MethodProphecy` [#430](https://github.com/phpspec/prophecy/pull/430).
+* [added] Deps: support for phpunit 8.0 *(with few warnings till we drop support for PHP5)* [#430](https://github.com/phpspec/prophecy/pull/430).
+* [added] Dev: convenience composer script `composer test`, will run phpspec and phpunit [#430](https://github.com/phpspec/prophecy/pull/430). 
+* [added] Dev: phpstan level 4 + `composer phpstan` convenience script [#430](https://github.com/phpspec/prophecy/pull/430).
```
stof: I would remove the PR link here. As all these are from the same PR anyway, they won't be useful (which is also a reason to split PRs btw)

CHANGES.md
```diff
@@ -1,3 +1,25 @@
...
+* [fixed] Minor typehint fixes [#430](https://github.com/phpspec/prophecy/pull/430).
+* [fixed] Minor: autoload-dev includes `Tests' and 'spec' namespaces [#430](https://github.com/phpspec/prophecy/pull/430).
+* [removed] Support support for PHP5.3, 5.4 and 5.5 (minimum version is now 5.6) [#430](https://github.com/phpspec/prophecy/pull/430).
+* [removed] Unused private method `CallCenter::formatExceptionMessage()` [#430](https://github.com/phpspec/prophecy/pull/430).
+* [removed] Deps: `phpdocumentor/reflection-docblock: ^2.0` and increased minimum version to ^3.1 [#430](https://github.com/phpspec/prophecy/pull/430).
+* [removed] Deps: `sebastian/recursion-context: ^1.0` and increased minimum version to ^2.0 [#430](https://github.com/phpspec/prophecy/pull/430).
+* [removed] Deps: `sebastian/comparator: ^1.1` and increased minimum version to ^1.2 [#430](https://github.com/phpspec/prophecy/pull/430).
+* [removed] Deps: removed `phpunit ^4.6` [#430](https://github.com/phpspec/prophecy/pull/430).
+* [q&a] Travis stages update [#430](https://github.com/phpspec/prophecy/pull/430).
```
stof: this change to the Travis config is useless in the changelog. People using the library don't care about such change.

CHANGES.md
stof: `autoload-dev` is irrelevant for users of the library

CHANGES.md
stof: the config runs at level 3, not at level 4

src/Prophecy/Argument.php
```diff
@@ -92,13 +92,13 @@ public static function cetera()
     /**
      * Checks that argument matches all tokens
      *
-     * @param mixed ... a list of tokens
+     * @param Token\TokenInterface|Token\ExactValueToken ...$tokenList a list of tokens
```
stof: that's wrong. It really accept anything, as any value not being a TokenInterface will be wrapped inside an ExactValueToken

src/Prophecy/Prophecy/MethodProphecy.php
```diff
@@ -92,9 +92,8 @@ public function __construct(ObjectProphecy $objectProphecy, $methodName, $argume
 
                     case 'Traversable':
                     case 'Generator':
-                        // Remove eval() when minimum version >=5.5
                         /** @var callable $generator */
```
stof: this inline phpdoc can be removed now, as any static analysis tool will already be able to know it now that there is no usage of `eval` anymore.

tests/Doubler/Generator/ClassMirrorTest.php
```diff
@@ -456,9 +457,21 @@ function it_doesnt_fail_when_method_is_extended_with_more_params()
...
+         */
+        $class = $this->prophesize(\ReflectionClass::class);
+
+        /**
+         * @var ObjectProphecy&\ReflectionMethod $method
+         */
+        $method = $this->prophesize(\ReflectionMethod::class);
+
+        /**
+         * @var ObjectProphecy&\ReflectionParameter $parameter
```
stof: this is a wrong type. This variable is **not** an instead of ReflectionParameter at all

.travis.yml
```diff
@@ -12,26 +12,97 @@ branches:
...
+      php: 7.3
+      script:
+        - vendor/bin/phpspec run -fpretty -v
+        - vendor/bin/phpunit
+
+    - stage: Tests
+      php: 7.2
+      script:
+          - vendor/bin/phpspec run -fpretty -v
+          - vendor/bin/phpunit
```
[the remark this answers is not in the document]
belgattitude: Actually yes :) Thxs

CHANGES.md
belgattitude: Yes my too enthusiastic fault. It's pathological ;) Sorry

CHANGES.md
belgattitude: Yes I tried  something. Just cleaned up a bit based on previous changelog, but feel free to pick up or drop things.

tests/Doubler/Generator/ClassMirrorTest.php
belgattitude: Oups, thank you, very complete review !

belgattitude: @stof, sorry, I admit I'm a bit lost here. And kept the original behavior till I understand.

> This variable is not an instead of ReflectionParameter at all

I'm using intersection between Prophecy and ReflectionParameter to help phpstan. But I think it is a ReflectionParameter... Did I miss something ?

src/Prophecy/Argument.php
belgattitude: Beautiful, It's fixed thanks

tests/Doubler/Generator/ClassMirrorTest.php
stof: no it is not. Try it with `instanceof`. Only the value returned by `->reveal()` will be a ReflectionParameter.

src/Prophecy/Promise/ThrowPromise.php
```diff
@@ -64,7 +64,7 @@ public function __construct($exception)
      * @param ObjectProphecy $object
      * @param MethodProphecy $method
      *
-     * @throws object
+     * @throws \Exception
```
stof: not true. It can also be used for Error.

belgattitude: Sorry for taking so long to come back. Not sure about throwing an object, but I've reverted the change.

tests/Doubler/Generator/ClassMirrorTest.php
belgattitude: You're right. Reverted.

ciaranmcnulty:
This will need rebasing - I'm going to do a release soon and can you please rebase after that?

TBH it's hard to review so many things at once

composer.json
```diff
@@ -18,16 +18,16 @@
     ],
 
     "require": {
-        "php":                               "^5.3|^7.0",
-        "phpdocumentor/reflection-docblock": "^2.0|^3.0.2|^4.0",
-        "sebastian/comparator":              "^1.1|^2.0|^3.0",
+        "php":                               "^5.6|^7.0",
+        "phpdocumentor/reflection-docblock": "^3.1|^4.0",
+        "sebastian/comparator":              "^1.2|^2.0|^3.0",
```
drupol:
```suggestion
        "sebastian/comparator":              "^1.2.3|^2.0|^3.0",
```

drupol: Discussed in #452

tests/Doubler/Generator/ClassMirrorTest.php
```diff
@@ -456,9 +457,9 @@ function it_doesnt_fail_when_method_is_extended_with_more_params()
     function it_changes_argument_names_if_they_are_varying()
     {
         // Use test doubles in this test, as arguments named ... in the Reflection API can only happen for internal classes
-        $class = $this->prophesize('ReflectionClass');
-        $method = $this->prophesize('ReflectionMethod');
-        $parameter = $this->prophesize('ReflectionParameter');
+        $class = $this->prophesize(\ReflectionClass::class);
```
drupol: Maybe we should add new use:

`use \ReflectionClass;`

and then remove the backslash from the code.

WDYT ?

ciaranmcnulty:
Closing because it's been unmergable for a while; if you do have energy to update please do re-open

PR #432, opened 2019-06-14
fix compatibility with PHP 7.4

smoench:


smoench:
@ciaranmcnulty you may have a I look at this PR and consider having another bugfix release. Currently it preventing me testing projects against PHP 7.4 without ignoring php deprecation.

Thanks in advance :-)

src/Prophecy/Doubler/Generator/ClassMirror.php
```diff
@@ -144,7 +144,7 @@ private function reflectMethodToNode(ReflectionMethod $method, Node\ClassNode $c
         }
 
         if (version_compare(PHP_VERSION, '7.0', '>=') && $method->hasReturnType()) {
-            $returnType = (string) $method->getReturnType();
+            $returnType = PHP_VERSION_ID >= 70100 ? $method->getReturnType()->getName() : (string) $method->getReturnType();
```
oleg-andreyev: `getReturnType` always returning `ReflectionType`, no needs to check php version, just `$method->getReturnType()->getName()`

smoench: `ReflectionType` has no `getName`-method. With PHP 7.1 `RefelctionNamedType` gets returned which has the `getName`-method.

jakzal: `$method->getReturnType()` might return `NULL` if method type is undefined.

smoench: In if / line before there is a `$method->hasReturnType()`

smoench:
Thank you, @stof!

theodorejb:
Can a new release be published that includes this fix?

PR #433, opened 2019-06-15
Avoid using deprecated ReflectionType::__toString()

morozov:
Prophecy uses `ReflectionType::__toString()` which is [deprecated](https://www.php.net/manual/en/reflectiontype.tostring.php) since PHP 7.1. If used with PHPUnit, combined with Symfony `DeprecationErrorHandler`, it fails the test suite on PHP 7.4.0alpha1 (see [example](https://travis-ci.org/doctrine/dbal/jobs/545999926#L609)).


jdreesen:
Duplicate of #432

PR #434, opened 2019-06-27
Fix specs for MethodProphecy

tkotosz:
Currently when running phpscpec for the MethodProphecy class then I get the following:

```
$ vendor/bin/phpspec r spec/Prophecy/Prophecy/MethodProphecySpec.php 
                                                                                 0
0 specs
0 examples 
0ms
```

It seems this happens because a new class is defined before the MethodProphecySpec class which seemingly not allowed.

I have moved that other class to the bottom of the file then the tests started to run but a few of them was failing so I fixed those in this PR.

after the fixes:
image

### DIFF
--- a/spec/Prophecy/Prophecy/ClassWithVoidTypeHintedMethods.php
+++ b/spec/Prophecy/Prophecy/ClassWithVoidTypeHintedMethods.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace spec\Prophecy\Prophecy;
+
+class ClassWithVoidTypeHintedMethods
+{
+    public function getVoid(): void
+    {
+    }
+}
--- a/spec/Prophecy/Prophecy/MethodProphecySpec.php
+++ b/spec/Prophecy/Prophecy/MethodProphecySpec.php
@@ -2,6 +2,7 @@
 
 namespace spec\Prophecy\Prophecy;
 
+use PhpSpec\Exception\Example\SkippingException;
 use PhpSpec\ObjectBehavior;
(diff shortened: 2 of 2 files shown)


tkotosz:
php 5 build fixed now, please let me now if there is a better way to fix it :)

spec/Prophecy/Prophecy/ClassWithVoidTypeHintedMethods.php5.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace spec\Prophecy\Prophecy;
+
+class ClassWithVoidTypeHintedMethods
+{
+    public function getVoid()
```
stof: Should rather be dropped entirely, as this class does not actually have void typehinted methods, and skipping this case in the specs on PHP 5.

tkotosz: will see how to create a double manually instead of asking in it as a method argument (like now in `its_constructor_records_default_callback_promise_for_return_type_hinted_methods`) otherwise the php5 build would fail since the `ClassWithVoidTypeHintedMethods` class would be missing

tkotosz: ok, managed to work around it

tkotosz:
@stof fixed the PR based on your feedback. Could you please have a look again?

tkotosz:
squashed commits to prep the PR for merge

PR #435, opened 2019-06-27
Add willYield feature to Method Prophecy

tkotosz:
Adds `willYield` functionality. Usage:
```
function it_should_do_something(MyCollaborator $myCollaborator) {
    // arrange
    $myCollaborator->generate()->willYield(['foo', 'bar']);
    $this->beConstructedWith($myCollaborator);

    // act
    $result = $this->doSomething();

    // assert
    $result->shouldBe('foo/bar');
}
```

Depends on https://github.com/phpspec/prophecy/pull/434 (will rebase this PR once the other one merged)

### DIFF
--- a/spec/Prophecy/Prophecy/MethodProphecySpec.php
+++ b/spec/Prophecy/Prophecy/MethodProphecySpec.php
@@ -143,6 +143,42 @@ class MethodProphecySpec extends ObjectBehavior
         $this->getPromise()->shouldBeAnInstanceOf('Prophecy\Promise\ReturnPromise');
     }
 
+    function it_adds_CallbackPromise_during_willYield_call(ObjectProphecy $objectProphecy)
+    {
+        if (PHP_VERSION_ID < 50500) {
+            throw new SkippingException('Yield language feature was introduced in >=5.5');
+        }
+
+        $objectProphecy->addMethodProphecy($this)->willReturn(null);
+
+        $this->willYield(array('foo', 'bar'));
+        $this->getPromise()->shouldBeAnInstanceOf('Prophecy\Promise\CallbackPromise');
+    }
+
+    function it_yields_elements_configured_in_willYield(ObjectProphecy $objectProphecy)
+    {
+        if (PHP_VERSION_ID < 70000) {
+            throw new SkippingException('Yield language feature was introduced in >=5.5 but shouldYield matcher only available in >=7.0');
+        }
+
+        $objectProphecy->addMethodProphecy($this)->willReturn(null);
+
+        $this->willYield(array('foo', 'bar'));
+        $this->getPromise()->execute(array(), $objectProphecy, $this)->shouldYield(array('foo', 'bar'));
+    }
+
+    function it_yields_key_value_paris_configured_in_willYield(ObjectProphecy $objectProphecy)
+    {
+        if (PHP_VERSION_ID < 70000) {
+            throw new SkippingException('Yield language feature was introduced in >=5.5 but shouldYield matcher only available in >=7.0');
+        }
+
+        $objectProphecy->addMethodProphecy($this)->willReturn(null);
+
+        $this->willYield(array(10 => 'foo', 11 => 'bar'));
+        $this->getPromise()->execute(array(), $objectProphecy, $this)->shouldYield(array(10 => 'foo', 11 => 'bar'));
+    }
+
     function it_adds_ThrowPromise_during_willThrow_call(ObjectProphecy $objectProphecy)
     {
         $objectProphecy->addMethodProphecy($this)->willReturn(null);
--- a/src/Prophecy/Prophecy/MethodProphecy.php
+++ b/src/Prophecy/Prophecy/MethodProphecy.php
@@ -181,6 +181,40 @@ class MethodProphecy
     }
 
     /**
+     * @param array $items
+     *
+     * @return $this
+     *
+     * @throws \Prophecy\Exception\InvalidArgumentException
+     */
+    public function willYield($items)
+    {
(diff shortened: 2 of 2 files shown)


Taluu:
How about `willYield($value1, $value2, ...)` instead of `willYield([$value1, $value2, ...])` ?

tkotosz:
The `shouldYield` expects array if I remember correctly, so I thought best to expect array here as well.

antonfries:
Is this already merged to master or in near future? I need exactly this in one of my bigger projects.

spec/Prophecy/Prophecy/ClassWithVoidTypeHintedMethods.php5.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace spec\Prophecy\Prophecy;
+
+class ClassWithVoidTypeHintedMethods
```
stof: this does not make sense, as it does not have any void typehint. We should not have this class definition (and we should skip the void-typehint case when running the specs on PHP 5)

spec/Prophecy/Prophecy/MethodProphecySpec.php
```diff
@@ -76,7 +77,7 @@ function its_constructor_records_default_callback_promise_for_return_type_hinted
             $objectProphecy->reveal()->willReturn($subject);
 
             $this->beConstructedWith($objectProphecy, 'getVoid');
-            $this->getPromise()->shouldBeAnInstanceOf(CallbackPromise::class);
+            $this->getPromise()->shouldBeAnInstanceOf('Prophecy\Promise\CallbackPromise');
```
stof: should be reverted

spec/Prophecy/Prophecy/MethodProphecySpec.php
```diff
@@ -133,6 +134,44 @@ function it_adds_ReturnPromise_during_willReturn_call($objectProphecy)
...
+            $this->willYield(array('foo', 'bar'));
+            $this->getPromise()->shouldBeAnInstanceOf('Prophecy\Promise\CallbackPromise');
+        }
+    }
+
+    function it_yields_elements_configured_in_willYield($objectProphecy)
+    {
+        // yield language feature only introduced in >=5.5
+        // but shouldYield matcher only available in >=7.0
+        if (version_compare(PHP_VERSION, '7.0', '>='))
```
stof: version comparison should use `\PHP_VERSION_ID >= 70000` (same in other places).

spec/Prophecy/Prophecy/MethodProphecySpec.php
```diff
@@ -133,6 +134,44 @@ function it_adds_ReturnPromise_during_willReturn_call($objectProphecy)
         $this->getPromise()->shouldBeAnInstanceOf('Prophecy\Promise\ReturnPromise');
     }
 
+    function it_adds_CallbackPromise_during_willYield_call($objectProphecy)
+    {
+        // yield language feature only introduced in >=5.5
+        if (version_compare(PHP_VERSION, '5.5', '>='))
```
stof: for unsupported features, we should not turn the case into a no-op, but skip it (by throwing a `SkippingException`)

tkotosz: make sense, I was just following what was in the spec file originally

spec/Prophecy/Prophecy/MethodProphecySpec.php
tkotosz: `CallbackPromise::class` fails with parse error in php5 builds that's why I changed it to string like in other tests

stof: hmm, indeed. Our composer.json still says `PHP >= 5.3`. Can you submit these compat fixes on their own (they should be merged immediately)

tkotosz:
@stof FYI the test changes (fixing the build) are in a separate PR originally https://github.com/phpspec/prophecy/pull/434 (so I assume that one need to be reviewed and finalized) before this one can go forward

spec/Prophecy/Prophecy/MethodProphecySpec.php
tkotosz: ye already did these in a separate PR originally as I mentioned in the PR description. See: https://github.com/phpspec/prophecy/pull/434

stof: yeah sorry, I missed that part. I saw it after commenting.

tkotosz:
@antonfries In the meantime while this is not merged you can get around the issue by doing this:

Mock method to say it yields nothing:
{code}
$collaborator->generate()->will(function() {
    if (false) { // yield nothing
        yield;
     }
});
{code}

Mock method to yield some entries:
{code}
$collaborator->generate()->will(function() {
    yield MySuperItem::fromInt(1);
    yield MySuperItem::fromInt(2);
});
{code}

spec/Prophecy/Prophecy/MethodProphecySpec.php
tkotosz: no prob... will push fixes into the other PR :)

antonfries:
@tkotosz Thanks, that will help me for the first infrastructure parts!

tkotosz:
@stof rebased this PR from master now since the other PR has been merged :)

tkotosz:
@stof rebased this PR from master again, to apply the recent build fix

ciaranmcnulty:
This looks really useful, thanks

tkotosz:
Thanks for the merge @ciaranmcnulty :tada:

PR #442, opened 2019-09-04
Hotfix: case insensitive methods

### LINKED ISSUE #165: Method prophecies are case sensitive
``` php
$object->callUser();

.......

$object = $this->prophesize('Object');
$object->calluser()->shouldBeCalled();
```

This will fail and throw `UnexpectedCallException` because of `getMethodProphecies('calluser')` returning empty array

``` php
    //class ObjectProphecy
    public function getMethodProphecies($methodName = null)
    {
        ....
        if (!isset($this->methodProphecies[$methodName])) {
            return array();
        }
        ....
    }

    //class CallCenter
    $matches = array();
    foreach ($prophecy->getMethodProphecies($methodName) as $methodProphecy) {
        if (0 < $score = $methodProphecy->getArgumentsWildcard()->scoreArguments($arguments)) {
            $matches[] = array($score, $methodProphecy);
        }
    }

    if (!count($matches)) {
        throw $this->createUnexpectedCallException($prophecy, $methodName, $arguments);
    }
```

I am not suggesting to change the behaviour, just wondering if it is what you expect while PHP is case-insensitive.

Cheers

michalbundyra:
Fixes #165

Prophecy was case sensitive for methods, but PHP is not.

Because of that we have an issue in https://github.com/zendframework/zend-expressive-swoole/pull/70 as method name in the extension has been changed from `rawcontent` to `rawContent`.

### DIFF
--- a/src/Prophecy/Call/CallCenter.php
+++ b/src/Prophecy/Call/CallCenter.php
@@ -138,11 +138,11 @@ class CallCenter
      */
     public function findCalls($methodName, ArgumentsWildcard $wildcard)
     {
-        $lowerMethodName = strtolower($methodName);
+        $methodName = strtolower($methodName);
 
         return array_values(
-            array_filter($this->recordedCalls, function (Call $call) use ($lowerMethodName, $wildcard) {
-                return $lowerMethodName === strtolower($call->getMethodName())
+            array_filter($this->recordedCalls, function (Call $call) use ($methodName, $wildcard) {
+                return $methodName === strtolower($call->getMethodName())
                     && 0 < $call->getScore($wildcard)
                 ;
             })
--- a/src/Prophecy/Call/CallCenter.php
+++ b/src/Prophecy/Call/CallCenter.php
@@ -74,7 +74,7 @@ class CallCenter
         }
 
         // If no method prophecies defined, then it's a dummy, so we'll just return null
-        if ('__destruct' === $methodName || 0 == count($prophecy->getMethodProphecies())) {
+        if ('__destruct' === strtolower($methodName) || 0 == count($prophecy->getMethodProphecies())) {
             $this->recordedCalls[] = new Call($methodName, $arguments, null, null, $file, $line);
 
             return null;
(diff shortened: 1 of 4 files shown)


stof:
should we rather store MethodProphecy and Call objects with a lowercased method name, to avoid lowercasing them on each access (and changing `getMethodProphecies` from `O(1)` to `O(n)`) ?

webimpress:
@stof 
This was my first approach, but then we need to change it in many places and the final messages are not going to be that clear as before. I would need also modify tests when I start storing lowercase method names.

In messages we are gonna see not what we have provided but lowercase method, for example:

```
expected calls were:
- findpropecymethodcalls(...)
```
instead of:
```
- findProphecyMethodCalls(...)
```

that's why I've implemented "less efficient" solution.

I can push the alternative solution to the other branch so you can have a look and decide.

stof:
Well, we could have MethodProphecy and Call store the original case internally, usable in messages, but having `ObjectProphecy::$methodProphecies` be indexed by lowercase name for instance (to keep the `O(1)` access.

webimpress:
@stof Updated now. Technically it's BC Break as `getMethodProphecies` gives different result now than before. I guess you've expected that. That's why I've modified some tests.

ciaranmcnulty:
Does enabling case-insensitivity fit with our 'don't allow bad practices' philosophy?

webimpress:
@ciaranmcnulty This is PHP internals and because of that it's really hard to write tests with swoole extension (across different versions).

src/Prophecy/Call/CallCenter.php
```diff
@@ -74,7 +74,7 @@ public function makeCall(ObjectProphecy $prophecy, $methodName, array $arguments
         }
 
         // If no method prophecies defined, then it's a dummy, so we'll just return null
-        if ('__destruct' === $methodName || 0 == count($prophecy->getMethodProphecies())) {
+        if ('__destruct' === strtolower($methodName) || 0 == count($prophecy->getMethodProphecies())) {
```
Ocramius: Probably needs a test checking `__DeStRuCt`?

ciaranmcnulty:
Thanks, in retrospect I think it's good to mirror PHP's behaviour (although developers should be calling methods using the correct case)

Issue #268, opened 2016-04-08
Nesting level too deep - recursive dependency


stepanselyuk:
Hello,
I got error (after update to 1.6):

```
PHP Fatal error:  Nesting level too deep - recursive dependency? in /var/www/vhosts/trim-backend/vendor/phpspec/prophecy/src/Prophecy/Argument/Token/ExactValueToken.php on line 80
```
There is no error on v1.5.

stof:
Are you using the CachedDoubler or the normal doubler ?

toby-griffiths:
I'm having this same problem today.  Not sure which doubler I'm using, so assuming it's the default one?

What further info might you need to help get to the bottom of this?  I tried using Xdebug, but it's doing something odd.  When I break at line 80 where it reports the problem, there's not much of a call stack & when I step into that line, it immediately jumps to the error handler step.

Here's the call stack…

<img width="1090" alt="screen shot 2016-11-09 at 09 36 49" src="https://cloud.githubusercontent.com/assets/4817007/20133800/d249e01e-a660-11e6-99e7-76fbf41bb113.png">

toby-griffiths:
For the record, fixing the test, fixes the error 😃

Jean85:
Had the same issue in a unit test, randomly, which was really strange; I was able to fix it, I hope this will help someone in the future.

The test looked like this:

```php
public function test()
{
    $product = $this->getProduct(); // new Product()...

    $someClass = $this->prophesize(SomeClass::class);
    $someClass->someMethod($product)
        ->shouldBeCalledTimes(1)
        ->willReturn(false);

    $classUnderTest = new ClassUnderTest($someClass->reveal());

    $classUnderTest->methodUnderTest($product); // this will work

    $classUnderTest->methodUnderTest($this->getProduct()); // this will randomly fail
}
```

The issue will present itself only if the test is executed many times (i.e. with `--repeat 300`) or inside a large test suite. The underlying problem is that `$product` is a very large Doctrine entity, with loops due to relations and inverse relations.

The trigger of the bug is the fact that `->methodUnderTest($product)` does meet the expectation very fast, because it's the same instance; `->methodUnderTest($this->getProduct())` does not pass the same instance of the expectation, so Prophecy falls back to an `assertEquals` which triggers the nesting level limit.

Bottom line:
 * using `assertEquals` on a Doctrine entity with relations is very dangerous
[...]

PR #446, opened 2019-10-07
Fix nesting error when comparing objects

scroach:
When mocking method calls with complex objects (doctrine entities) I got PHP errors today: "PHP Fatal error:  Nesting level too deep - recursive dependency?"

This issue has already been discussed here: https://github.com/phpspec/prophecy/issues/268

The problem lies in the class method `ExactValueToken#scoreArgument`
When the ObjectComparator throws a `ComparisonFailure ` this method just ignores it - I don't see why because the phpunit comparator works quite well and seems to already cover recursion.

`ExactValueToken` then goes on and tries to compare the 2 objects with `$argument == $this->value` which causes a PHP internal nesting error - see also: https://stackoverflow.com/a/19097159/2424814

I've fixed this and added some tests.

### DIFF
--- a/src/Prophecy/Argument/Token/ExactValueToken.php
+++ b/src/Prophecy/Argument/Token/ExactValueToken.php
@@ -59,7 +59,9 @@ class ExactValueToken implements TokenInterface
             try {
                 $comparator->assertEquals($argument, $this->value);
                 return 10;
-            } catch (ComparisonFailure $failure) {}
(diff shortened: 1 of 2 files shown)


drupol:
This looks good to me.

ciaranmcnulty:
This looks good; it's not clear from the original commits why the Exception was ignored

PR #449, opened 2019-11-15
Automated native_function_invocation fixes

draco2003:
Reproduce with:
php php-cs-fixer --rules=native_function_invocation fix ./ --allow-risky=yes
(php-cs-fixer from : https://cs.symfony.com/)

Details:
https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/

### DIFF
--- a/spec/Prophecy/Argument/ArgumentsWildcardSpec.php
+++ b/spec/Prophecy/Argument/ArgumentsWildcardSpec.php
@@ -11,8 +11,8 @@ class ArgumentsWildcardSpec extends ObjectBehavior
     {
         $this->beConstructedWith(array(42, 'zet', $object));
 
-        $class = get_class($object->getWrappedObject());
-        $hash  = spl_object_hash($object->getWrappedObject());
+        $class = \get_class($object->getWrappedObject());
+        $hash  = \spl_object_hash($object->getWrappedObject());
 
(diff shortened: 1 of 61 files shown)


stof:
There is a meaningful performance improvement only for compiler-optimized functions, as PHP will cache the resolution of function names anyway (we are talking about a 3ns difference between `benchFqGlobalFunction` and `benchGlobalFunction` in the benchmark, while the standard deviations of these runs are 18ns and 21ns respectively, so the difference is not statistically relevant)
That's why the Symfony coding standards are to prefix the invocation only for compiler-optimized functions (php-cs-fixer maintains that list), as the extra `\` everywhere was considered as being less readable.

I suggest doing the same here.

draco2003:
Agree that php-cs-fixer should be basing it off the list here:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php#L353
Will look into why it was more aggressively adding them. 

Also other Libraries are opting for the use syntax instead:
use function is_array;

If this improvement is desired, do you have a preference of one over the other?

stof:
Well, if you run the fixer as `--rules=native_function_invocation`, it runs it with the default options, which are to fully-qualify all native functions. It don't think the CLI argument allows to specify options.

stof:
`use function is_array;` is a no-go due to our supported PHP versions.

Issue #451, opened 2019-11-29
Fix methods signatures issue with sebastian/comparator:^1.1


drupol:
Hi,

There is an issue in `sebastian/comparator` version `1.1.0` until version `1.2.2`.

`PHP Fatal error:  Declaration of SebastianBergmann\Comparator\DOMNodeComparator::assertEquals($expected, $actual, $delta = 0, $canonicalize = false, $ignoreCase = false) must be compatible with SebastianBergmann\Comparator\ObjectComparator::assertEquals($expected, $actual, $delta = 0, $canonicalize = false, $ignoreCase = false, array &$processed = Array) in /home/pol/dev/git/psrcas/vendor/sebastian/comparator/src/DOMNodeComparator.php on line 149`

The issue comes from version `1.1.0` of `sebastian/comparator`.

See those two files:

* https://github.com/sebastianbergmann/comparator/blob/c484a80f97573ab934e37826dba0135a3301b26a/src/DOMNodeComparator.php#L87
* https://github.com/sebastianbergmann/comparator/blob/c484a80f97573ab934e37826dba0135a3301b26a/src/ObjectComparator.php#L85

Their signature is different and when installing using composer flag `--prefer-lowest`, it fails.

That issue has been fixed in [version 1.2.3](https://github.com/sebastianbergmann/comparator/blob/1.2.3/src/DOMNodeComparator.php) ([see commit](https://github.com/sebastianbergmann/comparator/commit/9d9ede5a25938443d697a92551d031e83ed62856))

Therefore I propose to raise the minimal version to `1.2.3` instead of `1.1.0`.

Link to PR: #452

drupol:
There is another issue.

If we do this, then when downgrading to sebastian/comparator:`1.2.3`, this package has also an issue with the signature ([see ClosureComparator.php](https://github.com/phpspec/prophecy/blob/f6811d96d97bdf400077a0cc100ae56aa32b9203/src/Prophecy/Comparator/ClosureComparator.php#L30)).

I'll push a commit to fix this.

PR #452, opened 2019-11-29
Issue #451: Raise minimum version of sebastian/comparator.

drupol:
This PR:

* [x] Relates to #451

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     },
 
     "require-dev": {
-        "phpspec/phpspec": "^2.5|^3.2",
+        "phpspec/phpspec": "^3.2",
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
Why is this not an issue in prophecy’s CI?

drupol:
It does fail locally: https://pastebin.com/k5ZP4vpN

drupol:
See https://github.com/phpspec/prophecy/pull/453

ciaranmcnulty:
ok, I can't see any BC breaks in comparator 👍 

Could we allow 2.x too?

ciaranmcnulty:
I didn't realise that raising the phpspec version would drop out the old versions of PHP 🤦‍♂ 

Will revert in master

PR #453, opened 2019-11-29
Fix typo

drupol:
This pr relates to #452 .

### DIFF
--- a/spec/Prophecy/Prophecy/MethodProphecySpec.php
+++ b/spec/Prophecy/Prophecy/MethodProphecySpec.php
@@ -167,7 +167,7 @@ class MethodProphecySpec extends ObjectBehavior
         $this->getPromise()->execute(array(), $objectProphecy, $this)->shouldYield(array('foo', 'bar'));
     }
 
-    function it_yields_key_value_paris_configured_in_willYield(ObjectProphecy $objectProphecy)
+    function it_yields_key_value_pairs_configured_in_willYield(ObjectProphecy $objectProphecy)
     {
         if (PHP_VERSION_ID < 70000) {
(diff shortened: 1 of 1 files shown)


drupol:
@ciaranmcnulty Do you have any feedback on this? Should we keep tests with php < 5.6 ? Is it still required ?

drupol:
As PHPSpec 3 is in end of life, shouldn't we also update its version ?

spec/Prophecy/Promise/ThrowPromiseSpec.php
```diff
@@ -48,9 +48,9 @@ function it_throws_error_instances(ObjectProphecy $object, MethodProphecy $metho
             throw new SkippingException('The class Error, introduced in PHP 7, does not exist');
         }
 
-        $this->beConstructedWith($exc = new \Error('Error exception'));
+        $this->beConstructedWith(new \Error('Error exception'));
 
-        $this->shouldThrow($exc)->duringExecute(array(), $object, $method);
+        $this->shouldThrow(\Error::class)->duringExecute(array(), $object, $method);
```
ciaranmcnulty: Why is this changed?

drupol: I think this was unrelated, I reverted it.

ciaranmcnulty:
@drupol we can't update past phpspec 3 apparently because of BC breaks

drupol:
@ciaranmcnulty as long as we support unsupported PHP version (< 5.6), then it's impossible to test with lowest deps on Travis.
I will revisit this as soon as we will drop those from Travis.

PR #455, opened 2019-12-17
Update Travis matrix and build using low dependencies for each PHP versions.

drupol:


drupol:
So, to summarize:

When building with low deps on php > 5.6:

It fails because of [this](https://travis-ci.org/phpspec/prophecy/jobs/626164267#L993): `yields elements configured in willYield`

That issue is caused by the minimal version of PHP that we use in Composer:

`"phpspec/phpspec": "^2.5 || ^3.2",`

A solution might be to remove ^2.5 and use:

`"phpspec/phpspec": "^3.2 || ^4.3.4",`

But that also [fails](https://travis-ci.org/phpspec/prophecy/jobs/626169811#L262) with version > 7 because 3.2 introduce return type hinting and interfaces were updated.
PHP < 5.6 is also failing because PHPSpec ^3 only support PHP >= 5.6.

Then, if we update the methods to match the interface (https://github.com/phpspec/prophecy/pull/455/commits/1f6cf7eb7d16df161131da65c385f6adad394b58), then, it will fail for PHP <= 5.6 !
Why ? Because 5.6 doesn't support return type in methods.

I don't know what's the best move here.
According to me we should not test PHP version < 7.1 but I don't know if it's what the maintainer wants.
Any other idea is more than welcome to fix this "clustefunk" !

stof:
> According to me we should not test PHP version < 7.1 but I don't know if it's what the maintainer wants.

We must test any version that we support. That's the main rule.

But given that both PHPUnit and PhpSpec (which are the main 2 consumers of Prophecy) require PHP 7.2+ in their latest version, I think it could be fine to do a new minor version of Prophecy requiring it as well (we might want to release existing unreleased changes first).

What do you think @ciaranmcnulty ?

ciaranmcnulty:
@stof Yeah I've merged a bunch of stuff today so will cut a release later and we can revisit the bumping / whatever the policy is for this project on LTS and version numbers.

Wouldn't want to do anything that'll disrupt phpunit

stof:
@ciaranmcnulty As I said, PHPUnit itself requires PHP 7.2 already. So dropping support for older PHP versions in Prophecy won't disrupt them.

ciaranmcnulty:
@stof it's more about how we then handle bugs etc. in the previous versions - cutting a new major lets us patch 1.x too...

Depends on whether we're supporting those users in future

IMO dropping support without any announcement in advance will screw over a user base

stof:
@ciaranmcnulty cutting a new minor release does not prevent us to have a `1.10.x` branch if we need to provide additional patches for 1.10 users.

ciaranmcnulty:
Is this closable now or is it still necessary?

drupol:
Yes, thanks.

PR #459, opened 2019-12-20
Pin dists on travis and test on PHP 7.4 stable

GrahamCampbell:
Pinning the dists will avoid the travis config breaking in the future when they next change the default.


ciaranmcnulty:
Thanks

PR #460, opened 2019-12-21
Compare callables as an exact match if they are identical

### LINKED ISSUE #458: 1.10.0 weird problem with callables?
Hi, we just noticed some strange failing tests after upgrading to 1.10.0. We isolated the problem to this test case. 

```php
require __DIR__ . '/vendor/autoload.php';

class A {
    public function calc(callable $callable): int
    {
        return 0;
    }
}

$prophet = new \Prophecy\Prophet();

$callable = function() {};
$a = $prophet->prophesize(A::class);
$a->calc($callable)->willReturn(1);

var_dump($a->reveal()->calc($callable));
```

with 1.9.0, the output is `int(1)`
with 1.10.0 there's a fatal error

```
PHP Fatal error:  Uncaught TypeError: Return value of Double\A\P1::calc() must be of the type int, null returned in /Users/albert/Projects/backend-app/vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(49) : eval()'d code:6
Stack trace:
#0 /Users/albert/Projects/backend-app/test_prophecy.php(18): Double\A\P1->calc(Object(Closure))
#1 {main}
  thrown in /Users/albert/Projects/backend-app/vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(49) : eval()'d code on line 6

Fatal error: Uncaught TypeError: Return value of Double\A\P1::calc() must be of the type int, null returned in /Users/albert/Projects/backend-app/vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(49) : eval()'d code:6
Stack trace:
#0 /Users/albert/Projects/backend-app/test_prophecy.php(18): Double\A\P1->calc(Object(Closure))
#1 {main}
  thrown in /Users/albert/Projects/backend-app/vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(49) : eval()'d code on line 6
```

we're using PHP 7.3.13.

The problem seems to be related somehow to the callable, it does not happen if the function argument is a scalar or an object. Any ideas? Thx!!

ciaranmcnulty:
Our comparator has always treated callables as not-equal in all cases, however before #446 that error was swallowed and we fell back to other comparison methods.

This fixes #458 by making Closures match when they're identical

It will still be best practice to match on something like `Argument::is()`

### DIFF
--- a/spec/Prophecy/Comparator/ClosureComparatorSpec.php
+++ b/spec/Prophecy/Comparator/ClosureComparatorSpec.php
@@ -25,15 +25,15 @@ class ClosureComparatorSpec extends ObjectBehavior
         $this->accepts(function(){}, function(){})->shouldReturn(true);
     }
 
-    function it_asserts_that_all_closures_are_different()
+    function it_asserts_that_different_closures_are_different()
     {
         $this->shouldThrow()->duringAssertEquals(function(){}, function(){});
     }
 
-    function it_asserts_that_all_closures_are_different_even_if_its_the_same_closure()
+    function it_asserts_that_closures_are_equal_if_its_the_same_closure()
     {
         $closure = function(){};
 
-        $this->shouldThrow()->duringAssertEquals($closure, $closure);
+        $this->shouldNotThrow()->duringAssertEquals($closure, $closure);
     }
 }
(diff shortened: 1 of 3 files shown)


acasademont:
@ciaranmcnulty thanks a lot 👍

PR #462, opened 2019-12-23
Add Argument::in() and Argument::notIn()

viniciusalonso:
I was working in some exercises using TDD and I needed to write an example like this:

```php
$rule = $this->prophesize(MultipleThreeRule::class);
                                                                        
$rule->condition(Argument::that(function($arg) {
    return in_array($arg, [3, 6, 9, 12]);
}))->willReturn(true);
```
The idea is guarantee that a value is in array.

```php
$rule->condition(12); // returns true
``` 
Then, I implemented an easier way to get to same result:

```php
$rule = $this->prophesize(MultipleThreeRule::class);
                                                                        
$rule->condition(Argument::in([3, 6, 9, 12]))->willReturn(true);
```

And I also I implemented the opposite way, when value is not in array:

```php
$rule = $this->prophesize(MultipleThreeRule::class);
                                                                        
$rule->condition(Argument::notIn([3, 6, 9, 12]))->willReturn(false);
```

### DIFF
--- a/src/Prophecy/Argument/Token/InArrayToken.php
+++ b/src/Prophecy/Argument/Token/InArrayToken.php
@@ -19,13 +19,16 @@ namespace Prophecy\Argument\Token;
 class InArrayToken implements TokenInterface
 {
     private $token = array();
+    private $strict;
 
     /**
      * @param array $arguments tokens
+     * @param bool $strict
      */
-    public function __construct(array $arguments)
+    public function __construct(array $arguments, $strict = true)
     {
         $this->token = $arguments;
+        $this->strict = $strict;
     }
 
     /**
@@ -41,7 +44,7 @@ class InArrayToken implements TokenInterface
             return false;
         }
 
-        if (in_array($argument, $this->token)) {
+        if (\in_array($argument, $this->token, $this->strict)) {
             return 8;
         }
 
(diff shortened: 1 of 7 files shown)


src/Prophecy/Argument/Token/InArrayToken.php
```diff
@@ -0,0 +1,71 @@
...
+ * @author Vinícius Alonso <vba321@hotmail.com>
+ */
+class InArrayToken implements TokenInterface
+{
+    private $token = array();
+
+    /**
+     * @param array $arguments tokens
+     */
+    public function __construct(array $arguments)
```
Taluu: you could probably add a `$strict` argument (defaulting to `true` ?) so we can have an optimized call on `in_array`

viniciusalonso: It's interesting, what should be exactly the `$strict` role in that method? I'd like to understand better your idea.

Taluu: `in_array` have a 3rd parameter for that. If it's `true` and a `use function in_array` or `\in_array` is used, an optimized opcode will be used.

viniciusalonso: Good point! Let me improve this chunk of code, thanks. :)

ciaranmcnulty:
Thanks @viniciusalonso - looks good

PR #464, opened 2020-01-20
Make this compatible with next versions of sebastian/comparator and sebastian/recursion-context

sebastianbergmann:
The new versions will be released on February 7 and will be 100% compatible. In fact, there will likely not be any code changes. I am merely bumping the major version because of dropped support for PHP < 7.3.


sebastianbergmann:
Thanks!

PR #467, opened 2020-03-03
MagicCallPath: Fix error when @method can't be parsed.

shulard:
In some case, the `@method` tag can't be parsed and this will return an
`phpDocumentor\Reflection\DocBlock\Tags\InvalidTag` instance.

The problem is that the `InvalidTag` object doesn't have the `getMethodName` method so it leads to fatal errors during execution:

```
Call to undefined method phpDocumentor\Reflection\DocBlock\Tags\InvalidTag::getMethodName()
```

The problem was found when mocking the `\Fake\Generator` class from `fzaninotto/faker` with the `@method`:

```
@method mixed randomElement(array $array = array('a', 'b', 'c'))
```

Not sure the fix I added is the correct behaviour, the specs must be updated regarding that change for the travis build to pass.

### DIFF
--- a/spec/Prophecy/Doubler/ClassPatch/MagicCallPatchSpec.php
+++ b/spec/Prophecy/Doubler/ClassPatch/MagicCallPatchSpec.php
@@ -101,8 +101,6 @@ class MagicalApi
 }
 
 /**
- * @method void invalidMethodDefinition
- * @method void
  * @method
  */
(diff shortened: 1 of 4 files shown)


stof:
I created a different fix in #468 which fixes the implementation of the ClassTagRetriever instead of changing the signature in the interface (note that my reason to make a fix was also `Faker\Generator`).

PR #468, opened 2020-03-05
Fix the handling of invalid phpdoc tags

stof:
Such tags are represented by an InvalidTag instance rather than a Method tag, but the name might still be `method`. As we want to ignore invalid tags (that's the reason to catch exceptions), these tags are filtered out.

### DIFF
--- a/src/Prophecy/PhpDocumentor/ClassTagRetriever.php
+++ b/src/Prophecy/PhpDocumentor/ClassTagRetriever.php
@@ -44,7 +44,15 @@ final class ClassTagRetriever implements MethodTagRetrieverInterface
                 $this->contextFactory->createFromReflector($reflectionClass)
             );
 
-            return $phpdoc->getTagsByName('method');
+            $methods = array();
+
+            foreach ($phpdoc->getTagsByName('method') as $tag) {
+                if ($tag instanceof Method) {
+                    $methods[] = $tag;
+                }
+            }
+
+            return $methods;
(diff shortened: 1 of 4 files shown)


stof:
I caught this when working on the `nelmio/alice` testsuite, which tries to generate a double for `Faker\Generator`. And this class has some phpdoc that the library considers invalid (I copied 2 `@method` annotations of Faker in my fixture).

src/Prophecy/PhpDocumentor/ClassTagRetriever.php
```diff
@@ -44,7 +44,15 @@ public function getTagList(\ReflectionClass $reflectionClass)
                 $this->contextFactory->createFromReflector($reflectionClass)
             );
 
-            return $phpdoc->getTagsByName('method');
+            $methods = array();
+
+            foreach ($phpdoc->getTagsByName('method') as $tag) {
+                if ($tag instanceof Method) {
```
shulard: In multiple places, there is reference to a `LegacyMethod` object which represent `phpDocumentor\Reflection\DocBlock\Tag\MethodTag`.

Is it relevant to check this type here too ?

stof: no. This ClassTagRetriever is the implementation for version 3+ of the phpdocumentor library. LegacyMethod are produced by the LegacyClassTagRetriever when using version 2 of the library.

PR #469, opened 2020-04-13
Fix clone for double object

tkulka:
Hello, prophecy has problem with cloned double object.
Here is example:
```
class Foo
{
    private $object;

    public function __construct($object)
    {
        $this->object = clone $object;
    }

    public function getObject()
    {
        return $this->object;
    }
}
```
And spec tests:
```
class FooSpec extends ObjectBehavior
{
    public function it_compares_cloned_object()
    {
        $object = new \stdClass();
        $this->beConstructedWith($object);

        $this->getObject()->shouldBeLike($object);
    }

    public function it_compares_cloned_double_object(\stdClass $object)
    {
        $this->beConstructedWith($object);

        $this->getObject()->shouldBeLike($object);
    }
}
```
I expect the same behavior for two tests but first passed and second failed with error:
```
      - it compares cloned double object
      expected [obj:Double\stdClass\P1], but got [obj:Double\stdClass\P1].
      
      @@ -1,5 +1,5 @@
      -Double\stdClass\P1 Object &000000002e014a3e000000003f3f6471 (
      -    'objectProphecyClosure' => Closure Object &000000002e014a34000000003f3f6471 (
      -        0 => Closure Object &000000002e014a34000000003f3f6471
      +Double\stdClass\P1 Object &000000002e014a22000000003f3f6471 (
      +    'objectProphecyClosure' => Closure Object &000000002e014a3d000000003f3f6471 (
      +        0 => Closure Object &000000002e014a3d000000003f3f6471
           )
       )
```

Also I found the solution and I would like to shere with you.

### DIFF
--- a/spec/Prophecy/Argument/Token/ExactValueTokenSpec.php
+++ b/spec/Prophecy/Argument/Token/ExactValueTokenSpec.php
@@ -41,6 +41,14 @@ class ExactValueTokenSpec extends ObjectBehavior
         $this->scoreArgument($value2)->shouldReturn(10);
     }
 
+    function it_scores_10_if_value_is_an_double_object_and_equal_to_argument(\stdClass $value)
+    {
+        $value2 = clone $value->getWrappedObject();
+
+        $this->beConstructedWith($value);
+        $this->scoreArgument($value2)->shouldReturn(10);
+    }
+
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
Thanks for this @tkulka!

PR #473, opened 2020-06-07
Add method for incrementing a prediction's call times

krystalcode:
In some complex test cases with data providers I find the need to be able to increment the count of an existing call times prediciton. For example, I register the prediction, but if some condition passes I want to easily increment the count  instead of registering it again. Othewise I'd have to keep track of how many times it should be called and that makes writing the tests more difficult.

I have added a method `shouldBeCalledAddTimes($count)`; if there is an existing call times prediction it will add the given count to the prediction's count, otherwise it will create a new one.

Let me know if that sounds the right approach and I can take this further.

### DIFF
--- a/src/Prophecy/Prediction/CallTimesPrediction.php
+++ b/src/Prophecy/Prediction/CallTimesPrediction.php
@@ -43,6 +43,15 @@ class CallTimesPrediction implements PredictionInterface
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
This seems ok I guess; I'm not too clear what the use case is

Can you write some tests?

Issue #475, opened 2020-07-02
Drop older versions


ciaranmcnulty:
I started looking at PHP8 support and don't think it's possible to maintain support across 2 major versions.

I'm sold on upping the PHP version in the next minor, but what to? 7.0, 7.2 or 7.4?

@stof, opinion?

ciaranmcnulty:
Instinct says 7.4...

stof:
IMO, we need to keep support at least for the maintained PHP versions (meaning that going 7.4+ might be too early).

Then, the question is what each dropped version gives us (both in term of PHP features that become usable, and in terms of versions of dependencies).
For instance, from a PoV of PHP features, PHP 7.3 does not really give much benefit compared to PHP 7.2. The bigger features are in 7.4.

ciaranmcnulty:
@stof Yeah as you can see in #476 there aren't any version checks for 7.3 which means

Looking forward to supporting 8.0 as well, most important thing is that we shoulddrop any versions that don't have replacements for the deprecated APIs in 8.0. I think that's actually only 7.1

I guess I'm happy making it 7.2 and sticking to the PHP supported version. Shall I update the PR to be 7.2?

PR #476, opened 2020-07-02
Minimum PHP version of 7.2

ciaranmcnulty:
I took the opportunity to see which older versions of packages actually worked on PHP 7.3 and update too

I think this is an important step if we are going to fix issues with PHP8

Relates to #475


src/Prophecy/Call/CallCenter.php
```diff
@@ -62,16 +62,9 @@ public function __construct(StringUtil $util = null)
...
-            // Limit argument was introduced in PHP 5.4.0
-            $backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 3);
-        } elseif (defined('DEBUG_BACKTRACE_IGNORE_ARGS')) {
-            // DEBUG_BACKTRACE_IGNORE_ARGS was introduced in PHP 5.3.6
-            $backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
-        } else {
-            $backtrace = debug_backtrace();
-        }
+        // Limit backtrace to last 3 calls as we don't use the rest
+        // Limit argument was introduced in PHP 5.4.0
```
stof, line 66: this second line of comment can be dropped

tests/Doubler/Generator/ClassMirrorTest.php
```diff
@@ -261,10 +262,11 @@ public function it_marks_final_methods_as_unextendable()
 
     /**
      * @test
-     * @expectedException Prophecy\Exception\InvalidArgumentException
      */
     public function it_throws_an_exception_if_interface_provided_instead_of_class()
     {
+        $this->expectException(\Prophecy\Exception\InvalidArgumentException::class);
```
stof, line 268: I suggest moving that just before the call in which we expect it, which avoids accepting exceptions triggered by the setup logic.

stof: and this could use a use statement.

src/Prophecy/Prophecy/MethodProphecy.php
```diff
@@ -92,10 +92,7 @@ public function __construct(ObjectProphecy $objectProphecy, $methodName, $argume
 
                     case 'Traversable':
                     case 'Generator':
-                        // Remove eval() when minimum version >=5.5
-                        /** @var callable $generator */
-                        $generator = eval('return function () { yield; };');
-                        return $generator();
+                        return (function () { yield; })();
```
Taluu, line 95: we could use a static callable, nay ?

src/Prophecy/Prophecy/MethodProphecy.php
```diff
@@ -203,13 +200,11 @@ public function willYield($items)
             ));
         }
 
-        // Remove eval() when minimum version >=5.5
-        /** @var callable $generator */
-        $generator = eval('return function() use ($items) {
+        $generator =  function() use ($items) {
```
Taluu, line 203: same here

ciaranmcnulty: I don't think we can, as it is later bound to a class

ciaranmcnulty:
Not sure why Scrutinizer can't do the composer install - they suggest committing composer.lock to resolve it

src/Prophecy/Prophecy/MethodProphecy.php
Taluu: Oh didn't see that, so I'm trusting your judgement. ;)

PR #477, opened 2020-07-04
PHP 8 support

Ayesh:
Work in progress to bring PHP 8 support for prophecy. 

Majority of the errors were deprecation notices. I fixed the [optional-before-required errors](https://php.watch/versions/8.0/deprecate-required-param-after-optional) and deprecation of `ReflectionParameter` methods [`getClass`, `isCallable`, and `isArray`](https://php.watch/versions/8.0/deprecated-reflectionparameter-methods) with branching code that checks `getType` is available, and then uses it instead of the shorthand now-deprecated functions. 

I'm not completely sure about the fixes related to the new [`mixed` type](https://php.watch/versions/8.0/mixed-type). 

After these changes, there are only 4 errors/failures remaining from PHPUnit:

```
1) Tests\Prophecy\Doubler\ClassPatch\MagicCallPatchTest::it_supports_classes_with_invalid_tags
Method ReflectionParameter::getClass() is deprecated

prophecy/tests/Doubler/ClassPatch/MagicCallPatchTest.php:23

2) Tests\Prophecy\Doubler\Generator\ClassMirrorTest::it_marks_required_args_without_types_as_not_optional
Required parameter $arg_2 follows optional parameter $arg_1

prophecy\fixtures\WithArguments.php:7
prophecy/tests/Doubler/Generator/ClassMirrorTest.php:67

3) Tests\Prophecy\Doubler\Generator\ClassMirrorTest::it_doesnt_use_scalar_typehints
Error: Call to a member function getArguments() on null

prophecy/tests/Doubler/Generator/ClassMirrorTest.php:386

4) Tests\Prophecy\Doubler\Generator\ClassMirrorTest::it_changes_argument_names_if_they_are_varying
Error: Call to undefined method ReflectionUnionType::isBuiltin()

prophecy/tests/Doubler/Generator/ClassMirrorTest.php:480
```

I cannot get xdebug to work with PHP nightly builds, so these errors do not contain any backtraces. Any insights would be appreciated. 
I tried to skip some of the tests (such as `#2` above) with `@require PHP <= 8` annotations for the test because the deprecated behavior is exactly what's this test is asserting. 

`ReflectionParameter::export` method is deprecated, and removed in PHP 8, which causes `#3` above.

### DIFF
--- a/src/Prophecy/Doubler/Generator/ClassMirror.php
+++ b/src/Prophecy/Doubler/Generator/ClassMirror.php
@@ -221,14 +221,6 @@ class ClassMirror
             return $className;
         }
 
-        if (true === $parameter->isArray()) {
-            return 'array';
-        }
-
-        if (true === $parameter->isCallable()) {
-            return 'callable';
-        }
-
         if (true === $parameter->hasType()) {
             return $parameter->getType()->getName();
         }
--- a/src/Prophecy/Doubler/Generator/ClassMirror.php
+++ b/src/Prophecy/Doubler/Generator/ClassMirror.php
@@ -236,20 +236,14 @@ class ClassMirror
 
     private function getParameterClassName(ReflectionParameter $parameter)
     {
-        try {
-            $type = $parameter->getType();
-            if (!$type) {
-                return null;
-            }
-            if ($type instanceof ReflectionNamedType && !$type->isBuiltin()) {
-                return $type->getName();
-            }
-
+        $type = $parameter->getType();
+        if (!$type) {
             return null;
-        } catch (\ReflectionException $e) {
-            preg_match('/\[\s\<\w+?>\s([\w,\\\]+)/s', $parameter, $matches);
-
-            return isset($matches[1]) ? $matches[1] : null;
         }
+        if ($type instanceof ReflectionNamedType && !$type->isBuiltin()) {
+            return $type->getName();
+        }
+
+        return null;
     }
 }
--- a/src/Prophecy/Doubler/Generator/ClassMirror.php
+++ b/src/Prophecy/Doubler/Generator/ClassMirror.php
@@ -15,6 +15,7 @@ use Prophecy\Exception\InvalidArgumentException;
 use Prophecy\Exception\Doubler\ClassMirrorException;
 use ReflectionClass;
 use ReflectionMethod;
+use ReflectionNamedType;
 use ReflectionParameter;
 
 /**
@@ -236,7 +237,15 @@ class ClassMirror
     private function getParameterClassName(ReflectionParameter $parameter)
     {
         try {
-            return $parameter->getClass() ? $parameter->getClass()->getName() : null;
+            $type = $parameter->getType();
+            if (!$type) {
+                return null;
+            }
+            if ($type instanceof ReflectionNamedType && !$type->isBuiltin()) {
+                return $type->getName();
+            }
+
+            return null;
         } catch (\ReflectionException $e) {
             preg_match('/\[\s\<\w+?>\s([\w,\\\]+)/s', $parameter, $matches);
 
--- a/src/Prophecy/Doubler/Generator/ClassMirror.php
+++ b/src/Prophecy/Doubler/Generator/ClassMirror.php
@@ -38,14 +38,13 @@ class ClassMirror
     /**
      * Reflects provided arguments into class node.
      *
-     * @param ReflectionClass   $class
+     * @param ReflectionClass|null $class
      * @param ReflectionClass[] $interfaces
      *
      * @return Node\ClassNode
      *
-     * @throws \Prophecy\Exception\InvalidArgumentException
      */
-    public function reflect(ReflectionClass $class = null, array $interfaces)
+    public function reflect(?ReflectionClass $class, array $interfaces)
     {
         $node = new Node\ClassNode;
 
--- a/src/Prophecy/Doubler/Generator/Node/ArgumentNode.php
+++ b/src/Prophecy/Doubler/Generator/Node/ArgumentNode.php
@@ -92,7 +92,7 @@ class ArgumentNode
 
     public function isNullable()
     {
-        return $this->isNullable;
+        return $this->isNullable && $this->typeHint !== 'mixed';
     }
 
     public function setAsNullable($isNullable = true)
--- a/src/Prophecy/Doubler/Generator/TypeHintReference.php
+++ b/src/Prophecy/Doubler/Generator/TypeHintReference.php
@@ -22,6 +22,9 @@ final class TypeHintReference
             case 'object':
                 return true;
 
+            case 'mixed':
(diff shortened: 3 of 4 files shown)


Ayesh:
Related: phpspec/phpspec#1314

ciaranmcnulty:
I’d like to get #476 merged which would then simplify this a lot 👍

Ayesh:
That would he lovely! We can remove a lot of `void` workarounds and always assume `getType` is available. I'll wait for those PRs, and then rebase this PR after that. 

Thank you.

ciaranmcnulty:
Ok I merged it. I actually started a branch like this one but the complexity got too much for me so I did that other change :)

If you can rebase that looks good

For debugging in php8 I had similar issues and resorted to running in 7.4 for debug just to figure out where calls were coming from

ciaranmcnulty:
Maybe split out the changes that are to get the existing suite paying, from changes to support new features (eg mixed hint)

The latter would need additional tests etc

Ayesh:
Thanks a lot @ciaranmcnulty . I sent PR #478 to enable builds. We can either remove workarounds for PHP <7.2 first, and then work on PHP 8 fixes, or the other way around.

src/Prophecy/Doubler/Generator/ClassMirror.php
```diff
@@ -250,7 +250,19 @@ private function isNullable(ReflectionParameter $parameter)
...
-            return $parameter->getClass() ? $parameter->getClass()->getName() : null;
+            if (!method_exists($parameter, 'getType')) {
+                return $parameter->getClass() ? $parameter->getClass()->getName() : null;
+            }
+
+            $type = $parameter->getType();
+            if (!$type || $type->isBuiltin()) {
+                return null;
+            }
+            if (!method_exists($type, 'getName')) {
```
stof, line 261: I'd rather add a check for `instanceof ReflectionNamedType`, which will be needed anyway when adding the compatibility with union types.

src/Prophecy/Doubler/Generator/ClassMirror.php
```diff
@@ -222,11 +222,11 @@ private function getTypeHint(ReflectionParameter $parameter)
             return $className;
         }
 
-        if (true === $parameter->isArray()) {
+        if (version_compare(PHP_VERSION, '8', '<') && true === $parameter->isArray()) {
```
stof, line 225: I suggest using `\PHP_VERSION_ID < 80000` instead, which can be resolved at compile time.

src/Prophecy/Doubler/Generator/ClassMirror.php
```diff
@@ -222,11 +222,11 @@ private function getTypeHint(ReflectionParameter $parameter)
...
             return 'array';
         }
 
-        if (version_compare(PHP_VERSION, '5.4', '>=') && true === $parameter->isCallable()) {
+        if (true === $parameter->isCallable()) {
```
stof, line 229: shouldn't we have a version check here too according to the commit message ?

stof:
please also rebase this PR now that code for PHP < 7.2 has been removed.

Ayesh:
Thanks a lot for the great feedback. I will rebase/update code with new minimum PHP version in mind and other changes. 

I used `version_compare()` because it was already being used in other areas, but I think breaking that convention is worth the dead code elimination improvements.

Ayesh:
@stof I rebased and made some changes in the code, and with PHP 7.2 as the minimum, I could remove a lot of code. It would be great if you could take a look.

src/Prophecy/Doubler/Generator/ClassMirror.php
```diff
@@ -38,14 +39,13 @@ class ClassMirror
...
-     * @param ReflectionClass   $class
+     * @param ReflectionClass|null $class
      * @param ReflectionClass[] $interfaces
      *
      * @return Node\ClassNode
      *
-     * @throws \Prophecy\Exception\InvalidArgumentException
      */
-    public function reflect(ReflectionClass $class = null, array $interfaces)
+    public function reflect(?ReflectionClass $class, array $interfaces)
```
GrahamCampbell, line 48: This is a major breaking change because the class is non-final. Please revert.

src/Prophecy/Doubler/Generator/ClassMirror.php
```diff
@@ -245,7 +237,15 @@ private function isNullable(ReflectionParameter $parameter)
...
+            $type = $parameter->getType();
+            if (!$type) {
+                return null;
+            }
+            if ($type instanceof ReflectionNamedType && !$type->isBuiltin()) {
+                return $type->getName();
+            }
+
+            return null;
         } catch (\ReflectionException $e) {
```
GrahamCampbell, line 249: Don't need this.

src/Prophecy/Doubler/Generator/ClassMirror.php
ciaranmcnulty: Aha so that's the reason

src/Prophecy/Doubler/Generator/ClassMirror.php
ciaranmcnulty: This weirdly does not seem to cause any issue https://3v4l.org/roQB8 #lolphp

src/Prophecy/Doubler/Generator/ClassMirror.php
ciaranmcnulty: That's true @Ayesh the try/catch is because `getClass` can throw, `getName` does not

src/Prophecy/Doubler/Generator/ClassMirror.php
GrahamCampbell: There is a different, issue however. The method is public, but the parameter has changed from optional to required.

GrahamCampbell: Oh, it's because the 2nd param is required, so PHP treats them all as required, and null is specially implemented.

ciaranmcnulty: @GrahamCampbell did you delete a comment? It's not an issue for callers because they have no way of taking advantage of the optionality

GrahamCampbell: Yeh, I replaced with with the comment that the 1st param was not really optional anyway

GrahamCampbell:
I this this PR is ready for merge now. It gets the ball rolling at least, even if it is not complete (just like the phpspec PR).

ciaranmcnulty:
Thanks, @Ayesh!

src/Prophecy/Doubler/Generator/ClassMirror.php
stof: yeah (not that qualifying the call is important for that. Relying on constant fallback will delay it until runtime)

PR #478, opened 2020-07-05
Travis CI: Enable builds on PHP nightly

Ayesh:
Split from #477. 

Contains only updates to Travis builds, so we can get quicker feedback on changes. This will start builds on PHP 8, which will fail, but Travis config allows PHP 8 builds to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,12 +21,16 @@ matrix:
       dist: bionic
     - php: 7.4
       dist: bionic
+    - php: nightly
+      env: COMPOSER_FLAGS='--ignore-platform-reqs'
   fast_finish: true
+  allow_failures:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -21,12 +21,16 @@ matrix:
       dist: bionic
     - php: 7.4
       dist: bionic
+    - php: nightly
+      env: COMPOSER_FLAGS='--ignore-platform-reqs'
```
ciaranmcnulty, line 25: I would rather we do `composer config platform.php 7.4.0` as a step before the install - that way we won't ignore missing extensions

ciaranmcnulty:
Thanks for the contribution @Ayesh

PR #483, opened 2020-07-07
Use recommended PHP constraint

dereuromark:
```
phpspec/prophecy v1.10.3 requires php ^5.3|^7.0 -> your PHP version (8.0.0-dev) does not satisfy that requirement.
    - phpspec/prophecy 1.11.x-dev requires php ^7.2 -> your PHP version (8.0.0-dev) does not satisfy that requirement.
    - phpspec/prophecy 1.11.0 requires php ^7.2 -> your PHP version (8.0.0-dev) does not satisfy that requirement.
```

It is recommended to keep PHP constraint open, same also do the other frameworks like Symfony and Cake etc.
This allows to easier adjust the code to work with next major early on.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
     "name":         "phpspec/prophecy",
-    "description":  "Highly opinionated mocking framework for PHP 5.3+",
+    "description":  "Highly opinionated mocking framework for PHP",
     "keywords":     ["Mock", "Stub", "Dummy", "Double", "Fake", "Spy"],
     "homepage":     "https://github.com/phpspec/prophecy",
     "type":         "library",
@@ -18,7 +18,7 @@
     ],
 
     "require": {
-        "php":                               "^7.2",
+        "php":                               ">=7.2",
         "phpdocumentor/reflection-docblock": "^5.0",
         "sebastian/comparator":              "^3.0 || ^4.0",
         "doctrine/instantiator":             "^1.2",


Incognito:
+1.

This is blocking our ability to get ready for php8 because we can't install phpunit on the php8 beta.

ciaranmcnulty:
I don't understand why we would do this - prophecy does not support PHP8 yet. 

`composer require phpspec/prophecy --ignore-platform-req=php` can't you?

dereuromark:
@ciaranmcnulty You should be doing this for any vendor library.
There is no need to overprotect the community here. The default and recommended way to set env (PHP) constraints is using the >= char.
See also Symfony, cakephp and other libraries that show this de facto and proven default.

ciaranmcnulty:
> The default and recommended way to set env (PHP) constraints is using the >= char.

I disagree, we were stung very badly by `>=5.3` when `7.0` came out across a few projects. 

I'm simply not willing to take on the maintenance burden of tagging a version as working on 8.0 (and 9.0, 10.0) when we have no idea if it does yet.

dereuromark:
https://twitter.com/nicolasgrekas/status/1263023258938548225
:/

ciaranmcnulty:
I have a lot of respect for Nicolas but this is an area I disagree with him. That's ok though, it's different projects :)

PR #485, opened 2020-07-07
Handle "self" type hint

### LINKED ISSUE #484: Double has signatures that are incompatible with prophesized interface when using self
Observed this twice with 2 different classes already while working on https://github.com/sonata-project/SonataAdminBundle/pull/6186 :

`Fatal error: Declaration of Double\ItemInterface\P22::setParent(?Double\ItemInterface\P22 $parent = NULL): Knp\Menu\ItemInterface must be compatible with Knp\Menu\ItemInterface::setParent(?Knp\Menu\ItemInterface $parent = NULL): Knp\Menu\ItemInterface in /home/greg/dev/SonataAdminBundle/vendor/bin/.phpunit/phpunit-8.5.0-0/vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(49)`

`Fatal error: Declaration of Double\FormInterface\P30::setParent(?Double\FormInterface\P30 $parent = NULL) must be compatible with Symfony\Component\Form\FormInterface::setParent(?Symfony\Component\Form\FormInterface $parent = NULL) in /home/greg/dev/SonataAdminBundle/vendor/bin/.phpunit/phpunit-8.5.0-0/vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(49)`

In both cases, the interface references itself in the signature.

Possible next steps:

- [x] create a minimal reproducer:
```php
<?php

use Prophecy\Prophet;

require 'vendor/autoload.php';

interface SelfReferencing
{
    public function __invoke(self $self);
}

$selfReferencing = (new Prophet)->prophesize(SelfReferencing::class);
$selfReferencing->reveal();
```
- [x]  bisect that issue down to a particular commit :arrow_forward: bf3f063a739b4f856ec0e88c75b61cb3a60ffc89 from #477
- [x]  Contribute [a failing test](https://github.com/phpspec/prophecy/pull/485)
- [x]  Diagnose the issue: `self` is a special case: https://3v4l.org/CVWn1 and needs to be handled accordingly
- [x]  Contribute [a fix](https://github.com/phpspec/prophecy/pull/485)

greg0ire:
Using self in the implementing class means narrowing the type, and
result in an LSP violation (and a rightful crash from PHP).
Resolving self to the declaring type results in a compatible signature.

Fixes #484

### DIFF
--- a/src/Prophecy/Doubler/Generator/ClassMirror.php
+++ b/src/Prophecy/Doubler/Generator/ClassMirror.php
@@ -241,6 +241,10 @@ class ClassMirror
             return null;
         }
         if ($type instanceof ReflectionNamedType && !$type->isBuiltin()) {
+            if ($type->getName() === 'self') {
+                return $parameter->getDeclaringClass()->getName();
+            }
+
             return $type->getName();
         }
 
--- a/tests/Doubler/Generator/ClassMirrorTest.php
+++ b/tests/Doubler/Generator/ClassMirrorTest.php
@@ -458,6 +458,16 @@ class ClassMirrorTest extends TestCase
 
     /**
      * @test
+     * @doesNotPerformAssertions
+     */
+    function it_doesnt_fail_to_mock_self_referencing_interface()
+    {
+        $class = $this->prophesize('Fixtures\Prophecy\SelfReferencing');
+        $class->reveal();
+    }
+
+    /**
+     * @test
      */
     function it_changes_argument_names_if_they_are_varying()
     {
--- a/fixtures/SelfReferencing.php
+++ b/fixtures/SelfReferencing.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace Fixtures\Prophecy;
+
+interface SelfReferencing
+{
+    public function __invoke(self $self): self;
+}


tests/Doubler/Generator/ClassMirrorTest.php
```diff
@@ -456,6 +456,15 @@ function it_doesnt_fail_when_method_is_extended_with_more_params()
...
     }
 
+    /**
+     * @test
+     */
+    function it_doesnt_fail_to_mock_self_referencing_interface()
+    {
+        $class = $this->prophesize('Fixtures\Prophecy\SelfReferencing');
+        $class->reveal();
+    }
```
greg0ire, line 466: Not sure at all if this file is the right place for this, please advise.

tests/Doubler/Generator/ClassMirrorTest.php
```diff
@@ -456,6 +456,16 @@ function it_doesnt_fail_when_method_is_extended_with_more_params()
...
     }
 
+    /**
+     * @test
+     * @doesNotPerformAssertions
+     */
+    function it_doesnt_fail_to_mock_self_referencing_interface()
+    {
+        $class = $this->prophesize('Fixtures\Prophecy\SelfReferencing');
+        $class->reveal();
```
Ayesh, line 466: Is it possible to add a PHPUnit assertion here? PHPUnit will otherwise complain this is a risky test otherwise.

src/Prophecy/Doubler/Generator/ClassMirror.php
```diff
@@ -241,6 +241,9 @@ private function getParameterClassName(ReflectionParameter $parameter)
             return null;
         }
         if ($type instanceof ReflectionNamedType && !$type->isBuiltin()) {
+            if ($type->getName() === 'self') {
```
Ayesh, line 244: This makes me think, if we should cover `static` which is allowed in PHP 8 as well.

tests/Doubler/Generator/ClassMirrorTest.php
greg0ire: It won't actually, see the special annotation I added to avoid that.

src/Prophecy/Doubler/Generator/ClassMirror.php
GrahamCampbell: No, static is not an allowed param type. Only a return type.

greg0ire: Why does 8.0.0-alpha1 output the same syntax error as other versions? https://3v4l.org/NLh9C

greg0ire: Makes me think, I should check if my fix works for return types too.

tests/Doubler/Generator/ClassMirrorTest.php
Ayesh: Nice, thank you - I just noticed the new push with the annotation.

src/Prophecy/Doubler/Generator/ClassMirror.php
GrahamCampbell: Very recent build of PHP 8 confirms that it really is only for return types:

image

GrahamCampbell: (ironic branch name in the screenshot 😆... fixing php 8 for another project)

greg0ire: Looks like it does. As for `static` in a return type, it does not seem to require special handling: https://3v4l.org/KMWgc

Ayesh: This is totally not related to this PR/issue, but in hindsight, `ClassMirror::reflectMethodToNode()` method might need some additional handling.

Ayesh: @GrahamCampbell - Hihi, it fits either way 👍🏼😉

tests/Doubler/Generator/ClassMirrorTest.php
ciaranmcnulty: I think it's fine yes

ciaranmcnulty:
Wait what, getType() returns `self` rather than the actual type? FFS PHP

This seems ok to me - it could also have been done in a ClassPatch but they're not really meant for generic issues like this I suppose

ciaranmcnulty:
Thanks @greg0ire

GrahamCampbell:
> Wait what, getType() returns self rather than the actual type? FFS PHP

Yep, this is by design, and is the same reason it cannot raise a reflection exception. `getType` is a purely syntactic operation. It just reads the type as is from the parser.

stof:
We need to do the same for `parent`

GrahamCampbell:
`parent` is not a valid type?

PR #489, opened 2020-07-09
Skip/remove tests that don't apply in PHP8

ciaranmcnulty:


tests/Doubler/Generator/ClassMirrorTest.php
```diff
@@ -385,6 +386,10 @@ public function it_throws_an_exception_if_not_reflection_provided_as_interface()
      */
     public function it_doesnt_use_scalar_typehints()
```
ciaranmcnulty, line 387: @stof I'm not sure I'm clear on the intent of this test - it was something about not adding scalar type hints but I'm unclear when that would happen. Do you recall?

ciaranmcnulty:
With these skipped, the nightly tests pass

tests/Doubler/Generator/ClassMirrorTest.php
stof: no idea

ciaranmcnulty: Damn, safe to skip do you think?

It's failing because the internal types are now type hinted... but it's so unclear to me what behaviour is there

jakzal: This is where it was introduced: https://github.com/phpspec/prophecy/pull/109 ( https://github.com/phpspec/prophecy/commit/40ed2a35d4be8a0fa0efb8f403160acd664afc78)

Seems to be HHVM related.

ciaranmcnulty: Oh you're right. 'stop doing a thing it used to do with HHVM'

ciaranmcnulty: Thanks @jakzal

Issue #482, opened 2020-07-07
Cannot double a method with union type


ciaranmcnulty:
This is already breaking phpspec tests because some internal classes now have union types (e.g. ReflectionParameter now takes `int|string` in its constructor)

PR #490, opened 2020-07-09
Check for ReflectionNamedType before asking for the type

ciaranmcnulty:
This fails when presented with a union type (cc @Ayesh)

### DIFF
--- a/src/Prophecy/Doubler/Generator/ClassMirror.php
+++ b/src/Prophecy/Doubler/Generator/ClassMirror.php
@@ -222,7 +222,7 @@ class ClassMirror
             return $className;
         }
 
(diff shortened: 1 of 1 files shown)


stof:
this still does not support UnionType, as it would remove the typehint entirely (which will break the generated child class)

ciaranmcnulty:
Yes #482 for that right now; I was concentrating on getting the existing behaviours covered

ciaranmcnulty:
@stof Huh actually this is not breaking the child class for union types.  Because we're on 7.2 dropping the type hints doesn't cause an error (although we should ensure the child has the right hints of course)

PR #492, opened 2020-07-15
Add php8 support

Glancu:


Glancu:
There are still issues with the deprecated method ReflectionParameter::getClass() but i don't know, how to fix.

DonCallisto:
I think we need to update `phpdocumentor/reflection-docblock` to `5.1` (https://github.com/phpspec/prophecy/blob/master/composer.json#L22) and everything should be ok.

Glancu:
> I think we need to update `phpdocumentor/reflection-docblock` to `5.1` (https://github.com/phpspec/prophecy/blob/master/composer.json#L22) and everything should be ok.
> 
> Edit: I've just seen that `5.1` still has the same problem. The fix is currently only on `master`.

Nice, that's works :smiley:

composer.json
```diff
@@ -46,5 +46,10 @@
         "branch-alias": {
             "dev-master": "1.11.x-dev"
         }
+    },
+    "config": {
```
garak, line 50: Maybe it's safer to set this in travis configuration?

Glancu: Done :smiley:

DonCallisto:
LGTM. We only should remember to get rid of `5.0-dev` when Prophecy will have a new tagged version.
Otherwise we could ask for `5.1.1` for `reflection-docblock`.

stof:
What is this PR about, considering that #477 is already merged ?

garak:
@stof the PR you mentioned doesn't add php8 to require section in composer.json

composer.json
```diff
@@ -18,8 +18,8 @@
     ],
 
     "require": {
-        "php":                               "^7.2",
-        "phpdocumentor/reflection-docblock": "^5.0",
+        "php":                               "^7.2 || ^8.0",
+        "phpdocumentor/reflection-docblock": "5.x-dev",
```
GrahamCampbell, line 22: Change this to `"^5.2"`

.travis.yml
```diff
@@ -29,6 +29,9 @@ matrix:
   allow_failures:
     - php: nightly
 
+before_install:
+  - if [[ $TRAVIS_PHP_VERSION == 'nightly' ]]; then composer config platform.php 7.4.99; fi;
```
GrahamCampbell, line 33: This is not a proper solution. Instead, you should make the tests function on both PHPUnit 8.x and 9.x, otherwise this PR carries no value.

ciaranmcnulty:
Thanks for the effort but until #482 is fixed and feature freeze has happened, we don't support PHP8

Can you please revert the composer changes? The change to the test looks interesting

garak:
> Thanks for the effort but until #482 is fixed and feature freeze has happened, we don't support PHP8

Uh? Mentioned issue is not fixed and anyway it's not about php8 support.
I don't get mention to "feature freeze" either. This is a proposal to support php8 in master branch

.travis.yml
ciaranmcnulty: yeah I don't want to do this, we can wait for the other dependencies to tag appropriate versions. Please remove

composer.json
```diff
@@ -18,8 +18,8 @@
     ],
 
     "require": {
-        "php":                               "^7.2",
-        "phpdocumentor/reflection-docblock": "^5.0",
+        "php":                               "^7.2 || ^8.0",
```
ciaranmcnulty, line 21: Prophecy doesn't yet work with PHP 8 so please remove this - I'll change the composer.json when it's working across packages

composer.json
ciaranmcnulty: I see Graham's done a PR for this now it's tagged so take this line out

tests/Doubler/ClassPatch/MagicCallPatchTest.php
```diff
@@ -13,6 +13,10 @@ class MagicCallPatchTest extends TestCase
      */
     public function it_supports_classes_with_invalid_tags()
     {
+        if (PHP_VERSION_ID >= 80000) {
```
ciaranmcnulty, line 16: Why skip this? It seems to be passing on the nightly build, is there something logically wrong with it?

ciaranmcnulty: The skip message doesn't make sense to me in this context, and the test is passing in PHP 8 so I think it's an error?

ciaranmcnulty:
The config changes are something I'll do when prophecy is ready. I can't see the logic behind the skipped test - if you want to open a PR with just that in it and an explanation please do

composer.json
garak: What's the rationale of supporting php8 only when stable?
This is a testing library, it's used by many other libraries that are trying to check if php8 will work *before* its release

Glancu:
> The config changes are something I'll do when prophecy is ready. I can't see the logic behind the skipped test - if you want to open a PR with just that in it and an explanation please do

Thanks for reply. I fixed all problems, so we can try again with my code add support PHP 8.

composer.json
ciaranmcnulty: Because I don't want to say it works with PHP8 when it doesn't - it's perfectly possible to install it with PHP8 (we are doing this with phpspec right now)

PR #495, opened 2020-08-12
Support PHP8's Union and Mixed types

### LINKED ISSUE #482: Cannot double a method with union type

ciaranmcnulty:
Resolves #482

This pushes the types logic from the ArgumentNode and MethodNode (and the TypeHintReference) down into smaller value objects for ArgumentTypeNode and ReturnTypeNodes, and then expands it to support Unions

### DIFF
--- a/spec/Prophecy/Doubler/Generator/ClassCodeGeneratorSpec.php
+++ b/spec/Prophecy/Doubler/Generator/ClassCodeGeneratorSpec.php
@@ -287,6 +287,44 @@ PHP;
         $code->shouldBe($expected);
     }
 
+    function it_generates_proper_code_for_union_return_types
+    (
+        ClassNode $class,
+        MethodNode $method,
+        ArgumentNode $argument
+    )
+    {
+        $class->getParentClass()->willReturn('stdClass');
+        $class->getInterfaces()->willReturn([]);
+        $class->getProperties()->willReturn([]);
+        $class->getMethods()->willReturn(array($method));
+
+        $method->getName()->willReturn('foo');
+        $method->getVisibility()->willReturn('public');
+        $method->isStatic()->willReturn(false);
+        $method->getArguments()->willReturn([]);
+        $method->getReturnTypeNode()->willReturn(new ReturnTypeNode('int', 'string', 'null'));
+        $method->returnsReference()->willReturn(false);
+        $method->getCode()->willReturn('');
+
+        $code = $this->generate('CustomClass', $class);
+
+        $expected =<<<'PHP'
+namespace  {
+class CustomClass extends \stdClass implements  {
+
+public  function foo(): int|string|null {
+
+}
+
+}
+}
+PHP;
+        $expected = strtr($expected, array("\r\n" => "\n", "\r" => "\n"));
+
+        $code->shouldBe($expected);
+    }
+
     function it_generates_empty_class_for_empty_ClassNode(ClassNode $class)
     {
         $class->getParentClass()->willReturn('stdClass');
--- a/spec/Prophecy/Doubler/Generator/ClassCodeGeneratorSpec.php
+++ b/spec/Prophecy/Doubler/Generator/ClassCodeGeneratorSpec.php
@@ -6,6 +6,7 @@ use phpDocumentor\Reflection\DocBlock\Tags\Method;
 use PhpSpec\ObjectBehavior;
 use Prophecy\Argument;
 use Prophecy\Doubler\Generator\Node\ArgumentNode;
+use Prophecy\Doubler\Generator\Node\ArgumentTypeNode;
 use Prophecy\Doubler\Generator\Node\ClassNode;
 use Prophecy\Doubler\Generator\Node\MethodNode;
 use Prophecy\Doubler\Generator\Node\ReturnTypeNode;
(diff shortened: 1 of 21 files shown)


src/Prophecy/Doubler/Generator/Node/ReturnTypeNode.php
```diff
@@ -0,0 +1,99 @@
...
+            case 'float':
+            case 'int':
+            case 'string':
+            case 'iterable':
+            case 'object':
+            case 'void':
+            case 'null':
+                return $type;
+            case 'mixed':
+                return PHP_VERSION_ID < 80000 ? $this->prefixWithNsSeparator($type) : $type;
```
stof, line 48: the constant should be fully qualified to resolve the version check at compile time

src/Prophecy/Doubler/Generator/Node/ReturnTypeNode.php
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+
+    private function guardIsValidReturn()
+    {
+        if ($this->types == ['null'=>'null']) {
+            throw new DoubleException('Return type cannot be standalone null');
+        }
+
+        if (isset($this->types['void']) && count($this->types) != 1) {
```
stof, line 61: should use strict comparison

src/Prophecy/Doubler/Generator/Node/ReturnTypeNode.php
```diff
@@ -0,0 +1,99 @@
...
+            $this->types,
+            function(string $type) {
+                return $type != 'null';
+            }
+        ));
+    }
+
+    private function prefixWithNsSeparator(string $type): string
+    {
+        return '\\' . ltrim('\\' . $type, '\\');
```
stof, line 92: why adding a `\` on the left to then trim all leading backslashes ? this will do exactly the same than not adding it, but with more work.

src/Prophecy/Doubler/Generator/Node/ReturnTypeNode.php
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    public function canUseNullShorthand(): bool
+    {
+        return isset($this->types['null']) && count($this->types) <= 2;
+    }
+
+    public function getNonNullTypes(): array
+    {
+        return array_values(array_filter(
```
stof, line 82: this could be simplified. As types are also set as key, instead of looping over the array to filter it with a callback, you can do this:

```
$types = $this->types;
unset($types['null']);

return array_values($types);
```

This loops over the array once instead of twice.

src/Prophecy/Doubler/Generator/Node/MethodNode.php
```diff
@@ -109,53 +105,69 @@ public function getArguments()
...
+        else {
+            $this->returnTypeNode = new ReturnTypeNode(...$this->returnTypeNode->getNonNullTypes());
         }
-        $this->returnType = $this->typeHintReference->isBuiltInReturnTypeHint($type) ?
-            $type :
-            '\\' . ltrim($type, '\\');
     }
 
+    /**
+     * @deprecated
```
stof, line 146: I suggest adding `use getReturnTypeNode instead.`

src/Prophecy/Doubler/Generator/Node/MethodNode.php
```diff
@@ -109,53 +105,69 @@ public function getArguments()
...
+    }
+
+    /**
+     * @deprecated
+     * @param bool $bool
+     */
+    public function setNullableReturnType($bool = true)
+    {
+        if ($bool) {
+            $this->returnTypeNode = new ReturnTypeNode(...array_merge($this->returnTypeNode->getTypes(), ['null']));
```
stof, line 138: no need to merge the array before spreading this. You could use `new ReturnTypeNode('null', ...$this->returnTypeNode->getTypes()` as we don't really care about the order.

spec/Prophecy/Doubler/Generator/ClassCodeGeneratorSpec.php
```diff
@@ -36,48 +37,39 @@ function it_generates_proper_php_code_for_specific_ClassNode(
...
 
         $method4->getName()->willReturn('doSomething');
         $method4->getVisibility()->willReturn('public');
         $method4->returnsReference()->willReturn(false);
         $method4->isStatic()->willReturn(false);
         $method4->getArguments()->willReturn(array());
-        $method4->hasReturnType()->willReturn(true);
-        $method4->getReturnType()->willReturn('void');
-        $method4->hasNullableReturnType()->willReturn(false);
+        $method4->getReturnTypeNode()->willReturn(new ReturnTypenode('void'));
```
stof, line 64: should use `new ReturnTypeNode` to respect the original case (and avoid class loading issues on case-sensitive filesystems, if the class is not loaded yet)

src/Prophecy/Doubler/Generator/Node/ReturnTypeNode.php
ciaranmcnulty: Yes it's an error / bad copy paste of https://github.com/phpspec/prophecy/blob/master/src/Prophecy/Doubler/Generator/Node/MethodNode.php#L137

src/Prophecy/Doubler/Generator/Node/MethodNode.php
ciaranmcnulty: Oh that's a nice idea - didn't think to use the spread like that 👍

src/Prophecy/Doubler/Generator/Node/ReturnTypeNode.php
ciaranmcnulty: Hm we have these in a few places, should it always be fully qualified?

Jean85 (approved):

ciaranmcnulty:
I believe this is now feature-complete

dereuromark:
I ran into the same issue today with https://travis-ci.org/github/milesj/decoda/jobs/728420774 and PHP8 not being testable right now.

src/Prophecy/Prophecy/MethodProphecy.php
```diff
@@ -71,14 +74,42 @@ public function __construct(ObjectProphecy $objectProphecy, $methodName, $argume
...
+            elseif ($reflectionType instanceof ReflectionUnionType) {
+                $types = $reflectionType->getTypes();
+            }
+
+            $types = array_map(
+                function(ReflectionType $type) { return $type->getName(); },
+                $types
+            );
+
+            usort(
```
ciaranmcnulty, line 92: This logic is totally wrong - needs more tests

src/Prophecy/Prophecy/MethodProphecy.php
```diff
@@ -91,13 +91,30 @@ function(ReflectionType $type) { return $type->getName(); },
 
             usort(
                 $types,
-                function(string $type) {
-                    if(class_exists($type) || interface_exists($type)) {
+                static function(string $type1, string $type2) {
```
ciaranmcnulty, line 94: Here we:

 1. Prefer to return a scalar or object promise rather than a null (existing behaviour)
 2. Prefer to return an object over a scalar (on the basis that the typical `Obj|false` case is similar to the above)
 3. If there are more than one object type in a union, or no objects and more than one scalar, we are just picking one of them essentially at random

3 here is maybe problematic, or at least counter-intuitive. We could sort by alphabetical order or something, but it doesn't seem to make sense. We could in theory use the order of the union except that ReflectionUnionType::getTypes 'has no guaranteed order'

PR #498, opened 2020-09-07
Fix PHP 8 issue with Named Arguments in array_merge

sanmai:
For example, see https://github.com/composer/composer/pull/9076


sanmai:
Seems to be a false alarm. Sorry about that.

PR #500, opened 2020-09-17
Force a specific phpspec hash until it has a PHP8 release

ciaranmcnulty:
Pins to a known good PHP8-supporting phpspec

Also stops nightly build from being an allowed failure. If we merge this, #499 and #495 we should have a passing PHP8 build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,12 +22,8 @@ matrix:
     - php: 7.4
       dist: bionic
     - php: nightly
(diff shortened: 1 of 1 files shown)


stof:
why not forcing the latest dev-master instead ?

ciaranmcnulty:
@stof the changes are on 6.2 branch

ciaranmcnulty:
I'd hope this to be a temporary situation. Once Prophecy allows 8.0 (soon) then phpspec can release pretty rapidly I believe

PR #502, opened 2020-09-28
Problem with inherited self return types

giovannialbero1992:
I have this behavior with the new minor version, maybe is a bug.

When I do a prophecy of an extended class that inherits a method with `self` return type I've this error:
```
Fatal error: Declaration of Double\Fixtures\Prophecy\ClassExtendAbstractWithMethodWithReturnType\P6::test(?DateTimeInterface $test = NULL): Fixtures\Prophecy\ClassExtendAbstractWithMethodWithReturnType must be compatible with Fixtures\Prophecy\AbstractBaseClassWithMethodWithReturnType::test(?DateTimeInterface $test): Fixtures\Prophecy\AbstractBaseClassWithMethodWithReturnType in /home/travis/build/giovannialbero1992/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(49) : eval()'d code on line 2
```
This error is raised into PHP versions <=7.3

### DIFF
--- a/fixtures/AbstractBaseClassWithMethodWithReturnType.php
+++ b/fixtures/AbstractBaseClassWithMethodWithReturnType.php
@@ -1,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Fixtures\Prophecy;
+
+abstract class AbstractBaseClassWithMethodWithReturnType implements EmptyInterface
+{
+    private $test;
+
+    public function test(?\DateTimeInterface $test): self
+    {
+        $this->test = $test;
+
+        return $this;
+    }
+}
--- a/fixtures/ClassExtendAbstractWithMethodWithReturnType.php
+++ b/fixtures/ClassExtendAbstractWithMethodWithReturnType.php
@@ -1,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Fixtures\Prophecy;
+
+class ClassExtendAbstractWithMethodWithReturnType extends AbstractBaseClassWithMethodWithReturnType
+{
+
+}
--- a/tests/ProphetTest.php
+++ b/tests/ProphetTest.php
@@ -1,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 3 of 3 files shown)


ciaranmcnulty:
Thanks for reporting, the reproducible test case is really great

lolli42:
Also reported in issue #503

alexander-schranz:
Can confirm this issue we have a [mock](https://github.com/sulu/sulu/blob/94fb1b2f5f6b7566c5f3765800567ed7bc0f0cfe/src/Sulu/Bundle/HttpCacheBundle/Tests/Unit/EventSubscriber/InvalidationSubscriberTest.php#L159) of our [StructureMetadata](https://github.com/sulu/sulu/blob/94fb1b2f5f6b7566c5f3765800567ed7bc0f0cfe/src/Sulu/Component/Content/Metadata/StructureMetadata.php) which implements [PropertiesMetadata](https://github.com/sulu/sulu/blob/94fb1b2f5f6b7566c5f3765800567ed7bc0f0cfe/src/Sulu/Component/Content/Metadata/PropertiesMetadata.php) and now fails with: 

> PHP Fatal error:  Declaration of Double\Sulu\Component\Content\Metadata\StructureMetadata\P23::setResource(string > $resource): Sulu\Component\Content\Metadata\StructureMetadata must be compatible with Sulu\Component\Content> \Metadata\PropertiesMetadata::setResource(string $resource): Sulu\Component\Content\Metadata\PropertiesMetadata in > C:\projects\sulu\vendor\phpspec\prophecy\src\Prophecy\Doubler\Generator\ClassCreator.php(49) : eval()'d code on line 2

ciaranmcnulty:
Yes we're generating the return type wrongly (we are basically behaving as if it's `static`) but this is allowed via PHP 7.4's type widening, which is why it's only failing on older versions.

Need to see how the old code coped with it (without a test case) and port the logic over.

ciaranmcnulty:
Thanks @giovannialbero1992 I've moved your commits to #504 and think it's fixed there

PR #507, opened 2020-10-24
imp: willYield handle Generator::getReturn()

camilledejoye:
The current implementation of `MethodProphecy::willYield()` does not allow to specify a return value for the mocked generator.

The proposal here is to add an optional parameter allowing to define the return value:
```php
$prophecy->createGenerator()->willYield(['foo', 'bar'], 'return value');
```


ciaranmcnulty:
Looks good, thank you!

Issue #472, opened 2020-05-28
Callcenter returns null on non registered method call


jeanlucc:
Hello,

First I do not often submit issues, so please accept my apologies if it is far from perfect.
This comment https://github.com/phpspec/prophecy/pull/441/files#r388191626 is perfectly true (as often with Stof) and thus creates an exception with an imprecise message.

Where I would like to see that there was no configured methodCall for these arguments with the list of provided arguments, I have a message about type mismatch.
This means to correct the error I have to look at the arguments with a debugger for instance.

I wonder if it is good to authorise a call that must return a value to fail later in case of `shouldHaveBeenCalled()` because it is not the job of prophecy to know/guess what to return.

I think an option is to return the good type thanks to reflection. Another option would be to check if the type of return is not `void` and throw the exception immediately which would mean users could not expect a `shouldHaveBeenCalled` to work for function with a return value without defining the correct return value themselves. There may be a lot of other better options but I only thought of these for now.

Could you please tell me where I made mistakes in my understandings. And if I'm correct about the problem how should we proceed to make this better?

PS : @everzet I'm a regular user of Prophecy and I want to thank you for this piece of software.

sstok:
I actually encountered this as well (and wanted to open an issue/pr about 😅 sorry).

A possible solution (suggested in Slack): In the case of a `TypeError` it would be possible to catch this and execute `checkUnexpectedCalls` so that at least reports why it failed?

jeanlucc:
Hello,

Thanks for your reply and no need to be sorry. I opened the issue because I thought it was a problem but I indeed implemented a work around that replace the TypeError Exception with the Exception of `checkUnexpectedCalls` so that I have a clear message.

But this needs to be implemented by everyone with this need (I think it might not be so many people) thus I raised the issue to know if it was an option to implement it directly in prophecy.

If your suggestion is for an implementation **in** Prophecy I do not see right now how to do it in a clean way and the advantages over the options I provided. (I do not know the implementation of Prophecy very well so it may be why I cannot see it right now).

ScreamingDev:
If I understood correctly, what you are trying to fulfill is an interface / return-type, which can already be done using Prophecy ...

```php
use Prophecy\Argument;

$proph = $this->prophesize(Some::class);
$proph->THE_METHOD()->yourRules(); // ...
$proph->THE_METHOD()->withArguments([Argument::cetera()])->willReturn('some-correct-type-here');
```

[...]

Issue #120, opened 2014-08-29
Spies do not register as allowed method calls when stubbing is specified


ciaranmcnulty:
Normally Spied-on methods are included in the list of expected calls. However, this breaks if stubbing takes place beforehand

Spec:

``` php
class SpiesExampleSpec extends ObjectBehavior
{
    function it_does_not_expect_spied_on_methods(\DateTime $time)
    {
        $time->getOffset()->willReturn(-1800);

        $this->modifyTimestamp($time);

        $time->setTimestamp(1234)->shouldHaveBeenCalled();
    }
}
```

Object:

``` php
class SpiesExample
{
    public function modifyTimestamp(\DateTime $time)
    {
        $time->getOffset();
        $time->setTimestamp(1234);
    }
}
```

Output:

```
  10  ! it does not expect spies
      method call:
        Double\DateTime\P1->setTimestamp(1234)
      was not expected.
      Expected calls are:
        - getOffset()
```

MPV:
I just ran in to this issue as well (I have a stub that I expect a method to be called on, but `shouldHaveBeenCalled()` gives the error that @ciaranmcnulty mentioned).

scaytrase:
Any news on this bug here in 2019?

jdreesen:
Can this be closed, now that #441 is merged?

PR #509, opened 2020-10-27
Add disable spy behavior

### LINKED ISSUE #508: [DX] Mocks should fail immediately
I was having some attrition while using Prophecy in my PHPUnit tests, and I track it down to mocks going further with unexpected calls.

I've opened https://github.com/phpspec/prophecy-phpunit/issues/27 thinking that the root issue was the migration from the old, deprecated PHPUnit integration to the new trait one, but I discovered that it wasn't, and instead it was https://github.com/phpspec/prophecy/issues/457 that, in 1.10, made this change.

That issue was closed by @ciaranmcnulty with:

> Yes, I think if faulty tests were wrongly passing before we don't need to consider it a break.

... but I'm here to ask for a revert, or at least a switch to enable the old behavior back.

**Why do I need this?**

I would like to go back to mock failing hard and fast because writing (or altering) tests with this new behavior is extremely frustrating: 
 * if the mocked class has a return type that doesn't allow null, I get a `\TypeError` (as pointed out by @stof in https://github.com/phpspec/prophecy/pull/441/files#r388191626)
 * if the mocked class has a return type that does allow null, I get a `null` that goes further in the SUT, often delaying the test failure and making it really obscure to track it down

In both those cases, it's often the case of getting the mock's expectation slightly wrong, but I have no way of discovering it: I always have to widen it up until it matches and I get a green test, and then go back, making the expectation narrower as desired.

Is it possible to get the old behaviour back? Or to have a way to opt-out from this new behaviour? I understand that #441 was not considered a BC break because code as is had to be correct anyway, but while changing code this is really frustrating!

Jean85:
Fix  #508

This PR is incomplete, I don't know how to fix the new specs.

### DIFF
--- a/spec/Prophecy/Call/CallCenterSpec.php
+++ b/spec/Prophecy/Call/CallCenterSpec.php
@@ -124,6 +124,33 @@ class CallCenterSpec extends ObjectBehavior
             ->shouldReturn('second');
     }
 
+    function createWithFailFast(bool $failFast)
+    {
+        $callCenter = new self();
+        $callCenter->makeUnexpectedCallsFailFast($failFast);
+        
+        return $callCenter;
+    }
+
+    function it_throws_exception_if_call_does_not_match_any_of_defined_method_prophecies_and_fail_fast_is_enabled(
+        $objectProphecy,
+        MethodProphecy $method,
+        ArgumentsWildcard $arguments
+    ) {
+        $method->getMethodName()->willReturn('getName');
+        $method->getArgumentsWildcard()->willReturn($arguments);
+        $arguments->scoreArguments(array('world', 'everything'))->willReturn(false);
+        $arguments->__toString()->willReturn('arg1, arg2');
+
+        $objectProphecy->getMethodProphecies()->willReturn(array('method1' => array($method)));
+        $objectProphecy->getMethodProphecies('getName')->willReturn(array($method));
+
+        $this->makeUnexpectedCallsFailFast(true);
+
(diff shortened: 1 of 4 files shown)


spec/Prophecy/Call/CallCenterSpec.php
```diff
@@ -124,6 +124,33 @@ function it_executes_promise_of_method_prophecy_that_matches_with_highest_score_
...
+    ) {
+        $method->getMethodName()->willReturn('getName');
+        $method->getArgumentsWildcard()->willReturn($arguments);
+        $arguments->scoreArguments(array('world', 'everything'))->willReturn(false);
+        $arguments->__toString()->willReturn('arg1, arg2');
+
+        $objectProphecy->getMethodProphecies()->willReturn(array('method1' => array($method)));
+        $objectProphecy->getMethodProphecies('getName')->willReturn(array($method));
+
+        $this->makeUnexpectedCallsFailFast(true);
```
Jean85, line 148: This is the test that got deleted in #120, I tried to revert it back but this line doesn't seem to work; I need to call that method as a new precondition to make it pass, how should I do it?

ciaranmcnulty:
How compatible is this with #518?

Jean85:
@ciaranmcnulty I've already commented there: https://github.com/phpspec/prophecy/pull/518#issuecomment-801091103

We're battling the same issue in opposite directions: Prophecy's behavior is unclear when the subject of the prophecy has return types declared on its methods. We should discuss how to proceed, probably in #472

PR #514, opened 2020-12-10
Fix the class attached to methodnotfoundexception

ciaranmcnulty:
The class was always being set to ObjectProphecy, which causes issues when trying to work out which class the method was not found on

An alternative would have been to use `$this->getProphecy()->reveal()` but that may cause side effects

Fixing this resolves https://github.com/phpspec/phpspec/issues/1350


stof:
this now passes a string rather than an object. Is it expected ?

ciaranmcnulty:
@stof the exception's constructor parameter is `object|string` and the accessor is called `getClassname`. There is one other place this is thrown and it's already being thrown as a string, so hopefully consumers expect it.

PR #515, opened 2020-12-17
Switch to Github Actions

ciaranmcnulty:


.github/workflows/build.yml
```diff
@@ -0,0 +1,37 @@
...
+
+    steps:
+      -   uses: actions/checkout@v2
+
+      -   name: Set up PHP
+          uses: shivammathur/setup-php@v2
+          with:
+            php-version: "${{ matrix.php }}"
+
+      -   name: Install dependencies
```
stof, line 30: to run both phpspec and phpunit even when phpspec fails (as done on Travis currently), here is the way to go:

```
-   name: Install dependencies
    run: COMPOSER_ROOT_VERSION=dev-master composer update ${{ matrix.composer-flags }}
    id: end-of-setup

-   name: Run tests (phpspec)
    run: bin/phpspec run --format=dot
    if: always() && end-of-setup.outcome == 'success'

-   name: Run tests (phpunit)
    run: ./vendor/bin/phpunit
    if: always() && end-of-setup.outcome == 'success'
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,37 @@
...
+          run: COMPOSER_ROOT_VERSION=dev-master composer update ${{ matrix.composer-flags }}
+
+      -   name: Run tests (phpspec)
+          run: bin/phpspec run --format=dot
```
stof, line 34: must be `vendor/bin/phpspec`

stof (changes requested):

.github/workflows/build.yml
```diff
@@ -0,0 +1,37 @@
...
+          - php: 7.2
+            composer-flags: "--prefer-lowest"
+
+    steps:
+      -   uses: actions/checkout@v2
+
+      -   name: Set up PHP
+          uses: shivammathur/setup-php@v2
+          with:
+            php-version: "${{ matrix.php }}"
```
jakzal, line 28: If you don't need coverage it can [be disabled](https://github.com/marketplace/actions/setup-php-action#disable-coverage) with `coverage: none`. This way xdebug won't affect performance of test runs.

.github/workflows/build.yml
ciaranmcnulty: Will this still fail the build if one fails?

PR #517, opened 2020-12-22
Add .github directory to .gitattributes export-ignore

Ayesh:
Add `.github` to `.gitattributes` with an `export-ignore` rule. This also aligns the `export-ignore` rules for more readability.


ciaranmcnulty:
Thanks!

stof:
To me, the alignment actually hurt maintenance: it means that adding a new line in the file might involve editing all lines of the file if the alignment has to change.

PR #525, opened 2021-03-10
Support the false pseudotype, which cannot be nullable or standalone

### LINKED ISSUE #524: Propehisizing a method with false in the return type leads to a fatal error in PHP8.0.2
Here are the issue details. If I try to mock this interface method it will fail with the below error  in PHP 8.0.2 > 

`PHP Fatal error:  Type declaration 'false' must be unqualified in 
/app/vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(49) : eval()'d code on line 5`

```
Interface Example:

interface TestInterface
{
    public function runTest(): array|false;
}


class InterfaceTest extends TestCase
{
    use ProphecyTrait;
    public function test_porpecy(): void
    {
        $mockTest = $this->prophesize(TestInterface::class);
        v$mockTest->runTest()->willReturn([]);

    }
}


```

### Discovery

We are currently upgrading PHP to version 8 and our unit test fails with the error described above. 

I think the main reason is multiple methods has return type like this 
```

 public  function getMultiByKey(string $server_key, array $keys, int $get_flags = 0): array|\false {
}
```

**false** is a keyword, not a global name so it cannot have  **\**

ciaranmcnulty:
Fixes #524

### DIFF
--- a/spec/Prophecy/Doubler/Generator/Node/ArgumentTypeNodeSpec.php
+++ b/spec/Prophecy/Doubler/Generator/Node/ArgumentTypeNodeSpec.php
@@ -84,4 +84,29 @@ class ArgumentTypeNodeSpec extends ObjectBehavior
             $this->shouldThrow(DoubleException::class)->duringInstantiation();
         }
     }
+
+    function it_does_not_prefix_false()
+    {
+        $this->beConstructedWith('false', 'array');
+
+        $this->getTypes()->shouldReturn(['false', 'array']);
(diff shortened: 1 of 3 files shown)


pardeep26k (approved):

pardeep26k:
> Fixes #524

@ciaranmcnulty  any idea when we merge and release this ?

jakzal (approved):

PR #526, opened 2021-03-17
Revert #441: 

### LINKED ISSUE #472: Callcenter returns null on non registered method call

Jean85:
Fixes #472, declares a new major bump to 2.0.

As suggested by @stof in https://github.com/phpspec/prophecy/issues/472#issuecomment-801142387, this is an attempt to revert #441 and go back to the old behavior, with a fast bailout when the prophecy hits an unregistered method call.
This avoids returning `null` in those cases, since it would probably lead to `\TypeError`s or other unpredictable behaviours, as originally pointed out in https://github.com/phpspec/prophecy/pull/441/files#r388191626.


spec/Prophecy/Prophecy/MethodProphecySpec.php
```diff
@@ -353,7 +353,7 @@ function it_records_even_failed_checked_predictions(
         $this->withArguments($arguments);
 
         try {
-            $this->callOnWrappedObject('shouldHave', array($prediction));
+          $this->callOnWrappedObject('shouldHave', array($prediction));
```
stof, line 356: this indentation change should be reverted

Jean85:
Ouch, we would need to declare PHPSpec compatible with Prophecy 2.0 in advance :cry: I'll revert the `dev-master` alias change to make the build happy.

ciaranmcnulty:
If we're going to bump the version number we should re-think the behaviour more comprehensively

What do you think of #528 ?

PR #529, opened 2021-05-14
Normalise Licence

exussum12:
Currently Github doesn't know what the licence is, this change allows Github to pick up the licence


ciaranmcnulty:
Thanks!

PR #534, opened 2021-07-30
support static closures in will and should

### LINKED ISSUE #506: static closure callback fails with CallbackPromise
Using static callbacks with CallbackPromise fails once Prophecy tries to bind the object to the callback closure

The following code:

$someProphecy->someMethod()->will(static fn () => 1);

Will fails with: 

> Cannot bind an instance to a static closure

Anonymous functions without the static keyword are working as expected

I don’t know how to check that a closure is static though

ntzm:
fixes #506

### DIFF
--- a/src/Prophecy/Prediction/CallbackPrediction.php
+++ b/src/Prophecy/Prediction/CallbackPrediction.php
@@ -16,6 +16,7 @@ use Prophecy\Prophecy\ObjectProphecy;
 use Prophecy\Prophecy\MethodProphecy;
 use Prophecy\Exception\InvalidArgumentException;
 use Closure;
+use ReflectionFunction;
 
 /**
  * Callback prediction.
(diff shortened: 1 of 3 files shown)


wouter90 (approved):

src/Prophecy/Prediction/CallbackPrediction.php
```diff
@@ -56,7 +57,7 @@ public function check(array $calls, ObjectProphecy $object, MethodProphecy $meth
     {
         $callback = $this->callback;
 
-        if ($callback instanceof Closure && method_exists('Closure', 'bind')) {
+        if ($callback instanceof Closure && method_exists('Closure', 'bind') && (new ReflectionFunction($callback))->getClosureThis() !== null) {
```
ntzm, line 60: Do we need the method_exists check any more? It was added in PHP 5.4 and Prophecy only supports 7.2+

ciaranmcnulty:
Thanks @ntzm !

PR #538, opened 2021-08-11
Support `never` return type

### LINKED ISSUE #537: Handle the `never` return type
We need to ensure we can correctly double classes with a `never` return

This may be as simple as our `void` handling, or could be more complex due to inheritance quirks

https://php.watch/versions/8.1/never-return-type

ciaranmcnulty:
Implementation closely mirrors how we treat `void`, closes #537

### DIFF
--- a/fixtures/NeverType.php
+++ b/fixtures/NeverType.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Fixtures\Prophecy;
+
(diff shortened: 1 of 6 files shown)


stof (approved):
Looks good to me

ciaranmcnulty:
Thanks @stof

PR #541, opened 2021-09-09
catch UnexpectedCallException

alexndlm:
We can safelly mark UnexpectedCallException as PredictionException


ciaranmcnulty:
This makes sense to me, can you please add a simple test? (maybe just checking the type is an instanceof)

alexndlm:
@ciaranmcnulty ping ping :)

Please, suggest where to add the test.

PR #542, opened 2021-09-09
Test that doubling an Enum causes a sensible error

### LINKED ISSUE #536: Doubling Enums
We will need to handle Enum types being doubled.

(If this is difficult  we could at least show a sensible 'not supported yet' message when it's attempted)

https://php.watch/versions/8.1/enums

ciaranmcnulty:
Just checks that it's a ClassMirrorException rather than some other fatal

Fixes #536

### DIFF
--- a/tests/Doubler/Generator/ClassMirrorTest.php
+++ b/tests/Doubler/Generator/ClassMirrorTest.php
@@ -603,4 +603,19 @@ class ClassMirrorTest extends TestCase
         $this->assertEquals(new ReturnTypeNode('never'), $methodNode->getReturnTypeNode());
 
     }
+
+    /**
+     * @test
+     */
+    public function it_can_not_double_an_enum()
+    {
+        if (PHP_VERSION_ID < 80100) {
+            $this->markTestSkipped('Enums are not supported in this PHP version');
+        }
+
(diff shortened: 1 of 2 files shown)


tests/Doubler/Generator/ClassMirrorTest.php
```diff
@@ -603,4 +603,19 @@ public function it_can_double_never_return_type()
...
+    /**
+     * @test
+     */
+    public function it_can_not_double_an_enum()
+    {
+        if (PHP_VERSION_ID < 80100) {
+            $this->markTestSkipped('Enums are not supported in this PHP version');
+        }
+
+        $this->expectException(ClassMirrorException::class);
```
Ocramius, line 616: Would a custom exception message make sense here?

Ocramius (changes requested):

tests/Doubler/Generator/ClassMirrorTest.php
ciaranmcnulty: The text is 'Could not reflect class <whatever>  as it is marked final` which I think is fine for now

Ocramius: Works 👍

Ocramius (approved):

Issue #535, opened 2021-08-10
Intersection types in generated code


ciaranmcnulty:
We need to ensure that if a class uses intersection return or argument type hints, that we generate the correct code when it's doubled.

Based on the work for union types this could be complex :/

https://php.watch/versions/8.1/intersection-types

kschatzle:
Somewhat looked into this. Going to throw a *wild* idea out there.

The slow incorporation of a code generation tool.

Incorporating a code generation tool into prophecy may allow for a faster development of the core mock'ing system currently in place. Code generation and parsing are a subject that has a depth and complexity that pushes the limit of casual understanding. Simply put, there's only so many hours in the day, and code generation vs unit testing seems to be two different buckets of work.

While looking to integrate intersection-type I immediately recognized the need to separate "UnionType" and "IntersectionType". Similar to how you guys have done with NodeTypeAbstract and children Return/Parameter.  I also see attributes of types "nullable" being useful. My underlying point is that I could personally see the codebase moving forward with a class hierarchy *similar* to other code generation libraries.

That begs the question, build it or utilize a library.

A code generation tool that has the attributes and associations built in are pretty close to what would be built regardless. Utilizing another library I *think* makes sense.

That said PHP-Parser is the tool that I would look into at the moment. The library is used in other contexts that require reading and writing of PHP code. I'm sure you guys could name them offhand (PHPStan, Rector, etc..)

I personally would not use every capability of PHP-Parser outright A certain amount of caution is called for with a user base as large as prophecy's. I could see a good first step in simply using PHP-Parser's [Node system](https://github.com/nikic/PHP-Parser/tree/master/lib/PhpParser/Node)  as a replacement for the current Prophecy's [Node system](https://github.com/phpspec/prophecy/tree/master/src/Prophecy/Doubler/Generator/Node).  The ClassMirror and ClassCodeGenerator would isolate PHP-Parser to ensure it can be removed if necessary at a future point.  The changes would also *not* use the parsing or code generation capability of PHP-Parser.  More time and resources may be needed to see if using those features even makes sense for Prophecy. Even if it does, a new major version may be necessary. Suffice to say that it seems like a larger undertaking when the immediate goals that I'd like to help with would be PHP version support.

I'm comparing that with the simplicity that you guys have with the current code generation. If you did not want to use PHP-Parser I would most likely mimic the hierarchy of their node systems combined with the current ClassMirror/Generation.

[...]

PR #543, opened 2021-09-09
Disallow intersection types with a nicer error message

ciaranmcnulty:
Does not really address #535 but makes doubling an intersection type produce a sensible error message, so unblocks an 8.1-supporting release


Ocramius (approved):
This is a good way to at least point users in the right direction ("not yet supported")

PR #551, opened 2022-01-24
Add PHP nightly support

gilbertsoft:


composer.json
```diff
@@ -18,7 +18,7 @@
     ],
 
     "require": {
-        "php":                               "^7.2 || ~8.0, <8.2",
+        "php":                               "^7.2 || ^8.0",
```
theofidry, line 21: I believe this was done on purpose. Some PHP libraries (and prophecy is one of them) are more sensitive to PHP API changes hence constraining it to a minor version is wiser.  I would instead recommend to use the `--ignore-platform-reqs=php+` for those cases were we want to test the upcoming yet not supported PHP versions

mvorisek, line 21:
```suggestion
        "php":                               ">=7.2 <8.3",
```

mvorisek:
It would be great if someone can merge this PR, PHP 8.2 alpha is already out.

.github/workflows/build.yml
```diff
@@ -15,14 +15,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: [7.2, 7.3, 7.4, 8.0]
+        php: [7.2, 7.3, 7.4, 8.0, 8.1, latest]
```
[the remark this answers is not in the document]
gilbertsoft: I don't think 8.2 should already be added here, should still be experimental. Will add it below though.

ciaranmcnulty:
I'm happy adding it explicitly to the matrix, we can add to composer.json at the RC stage I think? Prophecy will likely hold it back though in the real world

ciaranmcnulty:
We also need to go through the new features and see what breaks - for instance if you type hint the new types that are added in a test case - ideally someone could open an issue for each new feature that may be relevant and/or add a failing test and/or fix anything that breaks :)

composer.json
gilbertsoft: Reintroduced the upper limit again

gilbertsoft:
Added the first fixes.

Would be nice if someone could approve the tests....

stof:
Closing this as support for PHP 8.2 was added in 1.16 through other PRs.

PR #553, opened 2022-06-10
Support methods prefixed with underscore

### LINKED ISSUE #108: Cannot mock methods begining with underscore
See: https://github.com/phpspec/prophecy/blob/master/src/Prophecy/Doubler/Generator/ClassMirror.php#L106

What is the reason for this?

neclimdul:
Fixes #108

While it was a common convention to prefix "private" methods with an underscore
in earlier version of php, with support for method visibility this is no longer
common.

Additionally there are a number of exception such as the Soap library where
this convention did not hold and code was not mockable.


ciaranmcnulty:
I wonder if we'd need to do this in a major or not? I guess nobody's relying on 'methods that can't be mocked'?

borisson:
> I wonder if we'd need to do this in a major or not? I guess nobody's relying on 'methods that can't be mocked'?

I don't think this warrants a new major either, this behavior shouldn't make a change to anyone actually writing tests.
Only when they are doing some meta testing to see how many mockable methods a class has or something and even when typing it here, that gives me a headache ;)

PR #564, opened 2022-09-09
Initial compatability support for PHP8.2 for readonly classes.

kschatzle:
readonly classes will throw a ClassMirrorException until supported.

### DIFF
--- a/fixtures/ReadOnlyClass.php
+++ b/fixtures/ReadOnlyClass.php
@@ -1,0 +1,7 @@
+<?php
+
+namespace Fixtures\Prophecy;
+
+readonly class ReadOnlyClass
+{
+}
--- a/src/Prophecy/Doubler/Generator/ClassMirror.php
+++ b/src/Prophecy/Doubler/Generator/ClassMirror.php
@@ -98,6 +98,12 @@ class ClassMirror
             ), $class);
         }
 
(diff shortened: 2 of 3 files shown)


ciaranmcnulty:
What's the error currently when we try and extend readonly classes?

kschatzle:
> What's the error currently when we try and extend readonly classes?

```text
PHP Fatal error:  Non-readonly class Double\Fixtures\Prophecy\ReadOnlyClass\P1 cannot extend readonly class Fixtures\Prophecy\ReadOnlyClass in %s(%d) : eval()'d code on line %d
```

stof (approved):

gquemener:
I have pushed a patch to support doubling readonly class.
It's available through this PR: https://github.com/phpspec/prophecy/pull/571

Would there be a mean of communication (chat/mailing list) to coordinate efforts around supporting 8.2?
I'd gladly help, however don't want to create noise if you have already defined a strategy (first convert fatal error into proper exception, then support feature one by one, for instance).

ciaranmcnulty:
Thanks for the effort, I've merged #571 so this is not needed

ciaranmcnulty:
@gquemener on the subject of coordination I'd prefer Issues here

PR #565, opened 2022-09-09
Initial compatibility support for PHP8.2 for DNF

kschatzle:
DNF arguments and return types will throw ClassMirrorException's.


kschatzle:
Error without this patch is:
```
ParseError : syntax error, unexpected token "|", expecting variable
```

stof (approved):

PR #566, opened 2022-09-11
Support standalone and nullable types for true, false, and null.

kschatzle:


src/Prophecy/Doubler/Generator/Node/TypeNodeAbstract.php
```diff
@@ -61,6 +61,7 @@ protected function getRealType(string $type): string
             case 'callable':
             case 'bool':
             case 'false':
+            case 'true':
```
stof, line 64: true is not a builtin type before PHP 8.2, so it should not be in this branch. It needs a version check, as done for `mixed`

stof (changes requested):

kschatzle:
Question in general. Since true isn't a builtin type until PHP8.2, the code that is being mocked shouldn't have true as a standalone type. So how do choose what to prevent via versioning and what not?

src/Prophecy/Doubler/Generator/Node/TypeNodeAbstract.php
kschatzle: Done.

ciaranmcnulty:
@kschatzle not sure what the question means really? You have the PHP version checks in place that look correct, to me at least

Can you rebase please?

ciaranmcnulty:
ok I fixed the conflict and fixed an issue with true handling - thanks!

PR #572, opened 2022-10-17
Document arguments to willReturn

### LINKED ISSUE #552: func_get_args() use of willReturn() lacks proper annotation
Please add either an annotation or move to PHP variadics.

Otherwise static analysers will choke on every call of willReturn()

neclimdul:
Fixes #552

### DIFF
--- a/src/Prophecy/Prophecy/MethodProphecy.php
+++ b/src/Prophecy/Prophecy/MethodProphecy.php
@@ -241,11 +241,11 @@ class MethodProphecy
      *
      * @see \Prophecy\Promise\ReturnPromise
      *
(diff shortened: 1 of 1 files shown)


neclimdul:
variadic makes more sense to me but went with the annotation to match similar behavior in Argument::allOf

stof:
As the min supported version is now PHP 7.2, please use a variadic signature instead (the initial implementation was done with support for PHP 5.3).

stof:
@neclimdul any chance to update your PR ?

neclimdul:
@stof rebased and applied the variadic natively. Docs left for phpstan.

stof (approved):

PR #574, opened 2022-11-29
Add generic types for ObjectProphecy

### LINKED ISSUE #544: Feature Request: PHPDoc generics typehints
It would be amazing if we could typehint our prophecies with Generics phpdoc:

```

/**
 * @var ObjectProphecy<ClassBeingMocked>
 */
private $prophecy;
```

stof:
Closes #544

### DIFF
--- a/src/Prophecy/Prophecy/ObjectProphecy.php
+++ b/src/Prophecy/Prophecy/ObjectProphecy.php
@@ -26,6 +26,9 @@ use Prophecy\Exception\Prediction\PredictionException;
  * Object prophecy.
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * @template-covariant T of object
+ * @template-implements ProphecyInterface<T>
  */
 class ObjectProphecy implements ProphecyInterface
 {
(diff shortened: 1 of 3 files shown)


src/Prophecy/Prophecy/ObjectProphecy.php
```diff
@@ -80,6 +87,10 @@ public function willExtend($class)
      * @param string $interface
      *
      * @return $this
+     *
+     * @template U
+     * @phpstan-param class-string<U> $class
```
Wirone, line 92: @stof should be `$interface` here, not `$class`

ciaranmcnulty:
This looks cool but without tests/validation I find it hard to review ;) If we merge it 'wrong' will it break things for anyone?

Wirone:
If one is using PHPStan with `jangregor/phpstan-prophecy`, it probably will have priority over "native" annotations within library. I'm not sure about scenario where PHPStan is used without this plugin and there are errors in the baseline - then it depends on `reportUnmatchedIgnoredErrors` flag. Even if it's enabled, then all that has to be done is regenerating baseline.

stof:
@Wirone the goal is that a next version of `jangregor/phpstan-prophecy` removes its stub to rely on the built-in annotations instead. The extension would only have to deal with the magic API.

PR #575, opened 2022-11-29
Describe properly the return type of ObjectProphecy::getMethodProphecies

stof:


ciaranmcnulty:
It looks correct, should we add phpstan/psalm to the project? 😬

stof:
Well, analyzing our own code is a separate topic, but yes, it would be a good idea as a separate work.

PR #577, opened 2022-11-30
Add static analysis with phpstan

stof:


src/Prophecy/Prophecy/MethodProphecy.php
```diff
@@ -546,7 +555,7 @@ public function getMethodName()
     /**
      * Returns arguments wildcard.
      *
-     * @return Argument\ArgumentsWildcard
+     * @return Argument\ArgumentsWildcard|null
```
stof, line 558: This change should be backed out. The rest of Prophecy does not expect a `null` value there. And only specs for the class are using it with `null` as value (or with an array to be converted into an ArgumentsWildcard for that matter)

stof: https://github.com/phpspec/prophecy/pull/582 fixes the constructor to forbid passing `null`

stof:
@ciaranmcnulty I reached the max phpstan level, with only 15 errors added in the baseline. Most of them are related to case where we deal with string containing class names without defining their type as `class-string`, which is something that we can improve later. One of them is about `ReflectionClass::isReadOnly` being reported as undefined, which will go away once we run phpstan on PHP 8.2 (but as I have PHP 8.1 locally, it was easier that way for now). And 2 of them are false positives caused by https://github.com/phpstan/phpstan/issues/8439.
So this looks good enough for a merge IMO.

phpstan.neon
```diff
@@ -0,0 +1,8 @@
+parameters:
+    level: 9
+    inferPrivatePropertyTypeFromConstructor: true
```
stof, line 3: I'm using this to reduce the work in this PR as most our private properties don't have phpdoc. We can work later to define the types explicitly instead of relying on that feature (and I think Rector will be able to automate it for us) but I keep it separate.

PR #581, opened 2022-11-30
Use a private constant instead of a readonly static array

stof:


stof:
This fixes one of the issues reported by phpstan level 6 (types of static properties must be defined, while constants are inferred from the source code as they are not mutable)

PR #582, opened 2022-12-01
Remove support for instantiating a MethodProphecy without its arguments

stof:
The actual usage of the class in ObjectProphecy always specifies the argument.
Passing null would trigger fatal errors in multiple places in Prophecy due to the getter returning null instead of its expected type.


ciaranmcnulty:
Makes sense to me 👍

PR #585, opened 2023-01-11
Allow to work with PHPUnit 10

Jean85:
This draft of a PR was needed for me to work on PHPUnit 10 compat on another project, see https://github.com/facile-it/paraunit/pull/172

It allows two packages to the next major version (5), with no other code modifications since there are no BC impacting Prophecy.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
     "require": {
         "php":                               "^7.2 || 8.0.* || 8.1.* || 8.2.*",
         "phpdocumentor/reflection-docblock": "^5.2",
-        "sebastian/comparator":              "^3.0 || ^4.0",
+        "sebastian/comparator":              "^3.0 || ^4.0 || ^5.0",
         "doctrine/instantiator":             "^1.2 || ^2.0",
         "sebastian/recursion-context":       "^3.0 || ^4.0"
     },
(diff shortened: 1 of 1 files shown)


Jean85:
Rebased.

I left it as draft since both majors of both packages are not released yet.

Jean85:
@sebastianbergmann @stof  I've found a BC that's hitting this PR and it's going unnoticed due to the CI here not installing a dev dep: `SebastianBergmann\Comparator\Factory` if `final` and it's being extended in `Prophecy\Comparator\Factory`.

We could easily fix it with the decorator patter, but I don't know if it's to be considered a BC here (since no longer extending the class).

stof:
@Jean85 you need to update the specs of that Factory

Jean85:
@stof I didn't do it because, apart from the fact that I'm not skilled with PHPSpec, I'm not sure on how to proceed:
 * one test asserts that the class should extend the PHPUnit's one (should I change it to "not"? Should I expect it on the returned, decorated factory?)
 * The other one tests the behavior, which is changed...

stof:
I worked on preparing this work in https://github.com/phpspec/prophecy/pull/589

Jean85:
Closing and reopening to trigger a CI run, now that PHPUnit 10 is released.

Jean85:
This would require https://github.com/phpspec/phpspec/pull/1443 to test and work completely.

Jean85:
Let's follow the suggestion from https://github.com/phpspec/phpspec/pull/1443#issuecomment-1441442703 and test on an optional CI job requiring master.

PR #588, opened 2023-02-01
Add support for doctrine/instantiator 2.0

stof:
We are not affected by the BC break in it (which is most that some constants that were deprecated for public access are not actually private constants).

This keeps support for doctrine/instantiator 1.x because 2.0 requires PHP 8.1+ (and it makes it easier for projects to avoid dependency conflicts).


stof:
Note that for now the CI still uses doctrine/instantiator 1.x because of `phpspec/phpspec` that also depends on that library.

PR #591, opened 2023-02-10
allow recursion-context and comparator v5

### LINKED ISSUE #590: support sebastian/recursion-context v5
Hi,
I'm trying to install `prophecy-phpunit` alongside phpunit v10.

One issue that I am facing is that both phpunit and prophecy require `sebastian/recursion-context`, but phpunit 10 requires `^5.0` whereas prophecy requires `^3.0 || ^4.0` so they conflict.

brettmc:
PhpUnit 10 requires `sebastian/recursion-context` and `sebastian/comparator` v5. Allow these versions so that prophecy can be installed alongside PhpUnit 10 (eg for `prophecy-phpunit`).

Closes #590

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
     "require": {
         "php":                               "^7.2 || 8.0.* || 8.1.* || 8.2.*",
         "phpdocumentor/reflection-docblock": "^5.2",
-        "sebastian/comparator":              "^3.0 || ^4.0",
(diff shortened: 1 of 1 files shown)


stof:
See #585 for the in-progress work on that. This is on my plan for this weekend (hopefully) as it requires working on Prophecy 2.0 to remove the class deprecated in #589 (as sebastian/comparator v5 makes the Factory class final)

Issue #154, opened 2015-01-25
Unclear error message when MethodNotFound


ciaranmcnulty:
Currently when `Prophecy\Exception\Doubler\MethodNotFoundException` is thrown the message and exception includes the class of the double, not the interface or class being doubled.

This leads to messages like:
```
 method `Double\Person\P1::getName()` is not defined.
```

When it would be clearer for users to have something like:
```
 method `\Person::getName()` is not defined.
```

ciaranmcnulty:
I'm interested in catching and handling this error inside PhpSpec, but don't currently have a way to get the class/interface on which the method should be defined.

ciaranmcnulty:
Oh and anecdotally, this is something that is difficult for new learners; they find the message hard to interpret.

everzet:
Care to PR?

ciaranmcnulty:
It's on my list but thought I'd report first in case someone could pick it up :-)

flohw:
I encounter a similar problem recently. Let me know if I should create another issue.

When I mock a class but make a typo on the full class name (namespace and class name) then call my method to mock on my `ObjectProphecy` I get the same exception (`MethodNotFoundException`). I think it should throw another one like `ClassNotFoundException` before the `MethodNotFoundException`. The error could be understood more quickly.

nickedwards:
@flohw I just had this issue too. Scratching my head for a minute there, and this pointed me in the right direction, so thanks!

I agree that in this instance it should throw something like `ClassNorFoundException` instead. `MethodNotFoundException` didn't really make sense in that situation. However, I guess thats a minor issue.

jon-acker:
This is certainly a head-scratcher for newbies - has anyone picked it up? If not I can give it a go.

joachim-n:
I've had a poke around at this but I'm stumped.

At the point where the exception is thrown, we have access to the ObjectProphecy, but AFAICT to get from that the name of the class or interface for the prophecy would require getting them from its $lazyDouble property, which isn't accessible.

Would be useful to get some direction from maintainers on how to get this to work.

My first idea would be to implement __toString() on ObjectProphecy to output a human-friendly string representation of the prophecy for general debugging use.

stof:
Rather than `__toString` (which will enable casting to string in any context, which might lead to weird issue), I'd rather add a getter returning the expected class name

PR #597, opened 2023-04-17
feat: add control over class/interface existence before trying to prophesize it

Deuchnord:
Currently, if we try to prophesize a class or an interface that doesn't exist, the mock is created anyway, but with nothing inside. This behavior is disturbing, because then, when we try to set a method behavior inside, we get a "method not found" error, as said in https://github.com/phpspec/prophecy/issues/154#issuecomment-184572986:

```php
// The Person class does not exist
$mock = $this->prophesize(Person::class);
$mock->getName()->willReturn('Jérôme');
// Prophecy\Exception\Doubler\MethodNotFoundException: Method `Double\stdClass\P1::getName()` is not defined.
```

With this PR, trying to prophesize a class that does not exist will now throw a `ClassNotFoundException` with a comprehensive message that should help understanding what's actually happening.
This new behavior doesn't impact the creation of empty prophecy (i.e. calling `prophesize` without any argument). However, this might still be a BC break for people who use this method to mock inexistant classes on purpose.

### DIFF
--- a/spec/Prophecy/ProphetSpec.php
+++ b/spec/Prophecy/ProphetSpec.php
@@ -78,4 +78,13 @@ class ProphetSpec extends ObjectBehavior
     {
         $this->getDoubler()->shouldReturn($doubler);
     }
+
+    function it_throws_ClassNotFound_if_class_to_double_does_not_exist()
+    {
+        $this->shouldThrow('Prophecy\Exception\Doubler\ClassNotFoundException')
+            ->duringProphesize('This\ClassOrInterface\Does\Not\Exist');
+
+        $this->shouldNotThrow()
+            ->duringProphesize();
+    }
 }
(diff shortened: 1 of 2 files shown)


spec/Prophecy/ProphetSpec.php
```diff
@@ -78,4 +78,13 @@ function it_exposes_doubler_through_getter($doubler)
     {
         $this->getDoubler()->shouldReturn($doubler);
     }
+
+    function it_throws_ClassNotFound_if_class_to_double_does_not_exist()
+    {
+        $this->shouldThrow('Prophecy\Exception\Doubler\ClassNotFoundException')
+            ->duringProphesize('This\ClassOrInterface\Does\Not\Exist');
+
+        $this->shouldNotThrow()
```
stof, line 87: this should be a separate spec describing what it does instead of adding it in a spec saying that it throws.

stof (changes requested):

stof (approved):

jdreesen:
This is great! The current error message is very confusing.

PR #598, opened 2023-05-26
Do not allow null shorthand with standalone null type

Jean85:
I got an issue with `\SplFileObject::getChildren`, which in PHP 8.2 has a tentative return type of `null`. The code generator was creating a return type written as `): ? {`, creating a syntax error on the double. This fixes it.


Jean85:
CI build failure seems unrelated, probably due just to updated dependencies, especially PHPStan.
I'll try to address them in a separate PR.

PR #600, opened 2023-06-20
Fix the testsuite

stof:
As PHPUnit 10+ does not include a builtin Prophecy integration, we should not rely on it to write tests.

Thus, when writing tests for Prophecy itself, it is better if the test shows usages of the Prophecy API instead of hiding it in the PHPunit integration (which would also run additional code to verify predictions, which might make the test confusing).


stof:
I'm merging this. I will handle the phpstan failure in a separate PR.

Issue #607, opened 2023-08-22
PHP 8.3 support


vishalkhode1:
What's the plan to provide support for PHP 8.3, which is scheduled for release on November 23, 2023, with the first RC version set to be release on August 31, 2023.

stof:
It will depend on community contributions to identify which new features of PHP 8.3 are requiring changes in Prophecy to allow doubling them.
If you find incompatibilities, please report them as issues in the issue tracker (one issue per problem please, not a generic one mixing all compat issues as it makes it harder to track resolution). And then, if you have time to contribute a fix, it is even better.

To test things, you can use the ``--ignore-platform-req=php+` flag of composer to allow ignoring the upper bound of the PHP requirement to allow installing it on PHP 8.3 (that's exactly what this flag is about).

rajeshreeputra:
Created PR https://github.com/phpspec/prophecy/pull/608 attaching results:
<img width="674" alt="Screenshot 2023-08-24 at 12 34 16 PM" src="https://github.com/phpspec/prophecy/assets/19570710/7c478bdc-c7b8-4f3d-b823-1738ea9d08ad">

Also Created PR for phpspec/phpspec plugin for the same.

stof:
@rajeshreeputra having a green CI does not mean that we support PHP 8.3 though. It means we support code using older features when running on 8.3. Support for language features added in PHP 8.3 is not covered by the existing testsuite.

Ocramius:
I think `#[\Override]` may need some codegen adjustments here: 

![Selection_461](https://github.com/phpspec/prophecy/assets/154256/d5b39a3d-f66e-4d6f-a3e1-034da9a4a5ca)

https://wiki.php.net/rfc

stof:
Given that the `Override` attribute is not required when overriding a method (even when the parent method is itself using the attribute), I don't think it requires any code gen change.

Wirone:
From my perspective, what could be interesting from mocking library's perspective is:
- ability to mark anonymous class as readonly
- ability to overwrite class properties during cloning
- typed class constants maybe?

All of them (at least first 2) in terms of optimising generated code. But it's a wild guess only, I don't know about Prophecy's internal much.

stof:
Prophecy does not rely on anonymous classes (and currently does not support mocking readonly classes anyway as its internal implementation currently requires a mutable property in the mock class)

Jean85:
[...]

PR #608, opened 2023-08-24
PHP 8.3 compatibility checks.

rajeshreeputra:


.github/workflows/build.yml
```diff
@@ -2,7 +2,7 @@ name: Build
 
 on:
   push:
-    branches: [master]
+    branches: [master, php-8.3-compatibility]
```
stof, line 5: this should not be added there (or this should stay a branch in your fork without a PR in the main repo)

composer.json
```diff
@@ -18,7 +18,7 @@
     ],
 
     "require": {
-        "php":                               "^7.2 || 8.0.* || 8.1.* || 8.2.*",
+        "php":                               "^7.2 || 8.0.* || 8.1.* || 8.2.* || 8.3.*",
```
stof, line 21: Merging this is a no-go until PHP 8.3 is released as it would block us from doing releases. As said in #607, the way to try the PHP unreleased version is to use the `--ignore-platform-req=php+` option of composer.

.github/workflows/build.yml
```diff
@@ -49,7 +49,7 @@ jobs:
       - name: Set up PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: "8.2"
+          php-version: "8.3"
```
stof, line 52: I would not make phpstan run on an unstable PHP version as it increases the risk of having issues in phpstan itself.

.github/workflows/build.yml
rajeshreeputra: updated

.github/workflows/build.yml
stof: there is no reason to run phpstan on 2 different PHP versions IMO, except wasting compute resources for each CI build.

glo71317:
@stof Any plan to merge this PR because we are blocked due to this dependency in our code base.

stof:
@glo71317 my previous review comments are still relevant. We won't be releasing a version saying PHP 8.3 works right now. To try RC versions, the way to go is to use `--ignore-platform-req=php+`

composer.json
Wirone: @stof out of curiosity: why adding constraint for the next PHP version blocks the releases? 🤔

stof: Because for a mocking library, making a release that marks a PHP version as supported while it does not mock properly the code for the new version is a very bad idea.
As a mocking library, new language features often require adapting the library (and so are kind of BC breaks for us).

For devs wanting to try pre-release versions of 8.3, composer already provides exactly the feature you need for that (and testing newer PHP versions is the exact reason why this feature was introduced in composer, so it is the intended usage)

Wirone: Thank you very much for the explanation, totally makes sense 👍.

keradus: @stof, happy to share that 8.3 is now released ;) 
but I bet you already aware ;)

kylekatarnls:
It's released for a week now, this package is the only one still blocking to upgrade PHP without work-around at the moment some of the projects I work on. 🙏

kylekatarnls:
On another note, waiting for a PHP version to be stable-release to support it in a library sounds a ultra-drastic to me. Most of the libraries just tes in advance coming versions to ensure they are compatible, most of them also use range such as `^8.1` that just allow whatever <9 and depending on how deprecation are anticipated (for instance adding `8.4` to Github Actions now allow to test with the dev branch of PHP and detect incompatibility ahead before it can impact users of the library), updating can be actually smoother with a major range compatibility.

I don't say it's the only way, just raising alternatives in case it were not investigated in details.

Wirone:
@kylekatarnls PHP is _not_ SemVer, so using open constraints like `^8.1` is discouraged, especially for mocking library that needs to be up-to-date with language's syntax and features.

@stof but wouldn't it be possible to prepare internal test cases that just utilise PHP 8.3 features and verify if everything works? In Fixer we also don't use PHP 8.3 features (code is 7.4-compatible), but we have integration test that verifies if fixer properly fixes files with 8.3 syntax. Is it possible in Prophecy to prepare such tests?

stof:
@Wirone The issue is that since months that we said that this is the kind of checks that are needed to be able to know whether PHP 8.3 works fine (see https://github.com/phpspec/prophecy/issues/607), nobody in the community made such tests (or at least nobody reported the result to the repo if they tried it in separate codebase). It is not that it is not possible to do it. The issue is that it is not **done**.

PR #609, opened 2023-08-31
Add 'dev' keyword to composer.json to trigger `--dev` prompt

Ayesh:
Composer 2.4 and later can automatically suggest to use the `--dev` flag when someone runs `composer require phpspec/prophecy` without the `--dev` flag.

See [Get Composer to suggest dev packages to `require-dev`](https://php.watch/articles/composer-prompt-require-dev-dev-packages)


stof (approved):

Ayesh:
Thank you for the quick review and merge @stof 🙏

Issue #568, opened 2022-09-18
Method with a class argument and default value can't be mocked


bicpi:
Hi,
since 8.1, PHP supports the usage of the `new` operator in initializers (https://php.watch/rfcs/new_in_initializers).

Given this dummy class:

```php
class DummyClass {}
```

And an interface with a method using a `DummyClass` instance as default argument of a method:

```php
interface BarService
{
    public function doIt(DummyClass $myClass = new DummyClass()): void;
}
```

And a service using the interface as a dependency:

```php
class FooService
{
    public function __construct(
        private readonly BarService $bar
    ) {
    }

    public function execute(): void
    {
        $this->bar->doIt();
    }
}
```

The following test fails with a fatal PHP error in the `phpspec/prophecy` code:

```php
namespace Tests;

use PHPUnit\Framework\TestCase;
use Prophecy\PhpUnit\ProphecyTrait;

class DummyClass {...}
interface BarService {...}
class FooService {...}

class FooServiceTest extends TestCase
{
    use ProphecyTrait;

    /**
     * @test
     */
    public function it_fails(): void
    {
        $someService = $this->prophesize(BarService::class);
        $someService
            ->doIt()
            ->shouldBeCalled();

        $fooService = new FooService($someService->reveal());
        $fooService->execute();
    }
}
```

The error message is:

```
PHP Fatal error:  Constant expression contains invalid operations in /vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(49) : eval()'d code on line 5
```

Jean85:
I'm trying to debug this issue, and I cannot find any way to replicate the original behavior in the mock. The reflection gives you access only to `getDefaultValue()`, which returns the instantiated object, so it's impossible to produce written code that would get us the intended result. As of now, the generator uses `var_export` on it: https://github.com/phpspec/prophecy/blob/d4f454f7e1193933f04e6500de3e79191648ed0c/src/Prophecy/Doubler/Generator/ClassCodeGenerator.php#L112-L114

so this means calling `Class::__set_state(array(...))`, which generates the fatal. IMHO we should work around it for now and produce a `null` as a default value in place of the object.

stof:
Producing a `null` as default value is also broken, as it would change the method signature by making the argument nullable.

Jean85:
I proposed my change in #615

> Producing a `null` as default value is also broken, as it would change the method signature by making the argument nullable.

[...]

PR #615, opened 2024-01-24
Workaround for non-static default values for arguments

Jean85:
This is a workaround for #568.


stof:
See my comments on the issue explaining why this is wrong.

PR #618, opened 2024-03-26
Fix generation of enum default values with PHP 8.1

### LINKED ISSUE #567: Method with an enum argument and default value can't be mocked
Hi,
given this enum (PHP 8.1+):

```php
enum AppType
{
    case IOS;
    case ANDROID;
}
```

And an interface with a method using the enum as default argument of a method:

```php
interface BarService
{
    public function doIt(AppType $appType = AppType::ANDROID): void;
}
```

And a service using the interface as a dependency:

```php
class FooService
{
    public function __construct(
        readonly private BarService $bar
    ) {
    }

    public function execute(): void
    {
        $this->bar->doIt();
    }
}
```

The following test fails:

```php
namespace Tests;

use PHPUnit\Framework\TestCase;
use Prophecy\PhpUnit\ProphecyTrait;

enum AppType {...}
interface BarService {...}
class FooService {...}

class FooServiceTest extends TestCase
{
    use ProphecyTrait;

    /**
     * @test
     */
    public function it_fails(): void
    {
        $someService = $this->prophesize(BarService::class);
        $someService
            ->doIt()
            ->shouldBeCalled();

        $fooService = new FooService($someService->reveal());
        $fooService->execute();
    }
}
```

The error message is:

```
Error : Class "Double\BarService\Tests\AppType" not found
```

When passing the default value explicitly works though.

jdreesen:
Fixes #567

The problem is, that the generated default value is in the wrong namespace, because the leading `\` is missing. 

This seems only to be a problem in PHP 8.1, see: https://3v4l.org/3JPCM

### DIFF
--- a/src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
+++ b/src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
@@ -110,7 +110,13 @@ class ClassCodeGenerator
             $php .= '$'.$argument->getName();
 
             if ($argument->isOptional() && !$argument->isVariadic()) {
-                $php .= ' = '.var_export($argument->getDefault(), true);
+                $default = var_export($argument->getDefault(), true);
+
+                if ($argument->getDefault() instanceof \UnitEnum && 0 !== strpos($default, '\\'))  {
+                    $default = '\\'.$default;
+                }
+
+                $php .= ' = '.$default;
             }
 
             return $php;
--- a/src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
+++ b/src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
@@ -112,6 +112,7 @@ class ClassCodeGenerator
             if ($argument->isOptional() && !$argument->isVariadic()) {
                 $default = var_export($argument->getDefault(), true);
 
+                // This is necessary for PHP 8.1, as enum cases are exported without a leading slash in this version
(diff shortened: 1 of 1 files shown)


src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
```diff
@@ -110,7 +110,13 @@ private function generateArguments(array $arguments): array
             $php .= '$'.$argument->getName();
 
             if ($argument->isOptional() && !$argument->isVariadic()) {
-                $php .= ' = '.var_export($argument->getDefault(), true);
+                $default = var_export($argument->getDefault(), true);
+
+                if ($argument->getDefault() instanceof \UnitEnum && 0 !== strpos($default, '\\'))  {
```
stof, line 115: Can you add a comment saying this is for PHP 8.1, so that we can know this is dead code once we bump the min PHP version to something being at least PHP 8.2 (which is not planned soon, but will probably happen in a few years) ?

jdreesen:
```suggestion
                // This is necessary for PHP 8.1, as enum cases are exported without a leading slash in this version
                if ($argument->getDefault() instanceof \UnitEnum && 0 !== strpos($default, '\\'))  {
```

jdreesen:
CI failure is unrelated.

stof (approved):

jdreesen:
Would it be possible to get a release with this fix?

PR #621, opened 2024-03-30
Support mocking methods with a "static" return type

### LINKED ISSUE #620: Support for mocking methods with "static" return type
When attempting to mock a method with a `static` return type (see https://php.watch/versions/8.0/static-return-type), then you'll get this error:

```
Prophecy\Exception\Prophecy\MethodProphecyException: Cannot create a return value for the method as the type "static" is not supported. Configure an explicit return value instead.
```

Code:
```php
class Process implements \IteratorAggregate
{
    public function mustRun(?callable $callback = null, array $env = []): static
    {
        return $this;
    }
}
```

aik099:
The build failure is unrelated.

Closes #620

### DIFF
--- a/spec/Prophecy/Prophecy/MethodProphecySpec.php
+++ b/spec/Prophecy/Prophecy/MethodProphecySpec.php
@@ -567,6 +567,33 @@ class MethodProphecySpec extends ObjectBehavior
         $return->shouldImplement(ProphecySubjectInterface::class);
     }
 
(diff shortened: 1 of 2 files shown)


stof:
This Pr would end up doubling a double class (which is totally untested). I'd rather not do it.

aik099:
Good that you've noticed. I'm completely new to prophecy code (from the development side).

I'd better close this.

Issue #586, opened 2023-01-25
Mocking PHP 8.2 readonly classes generated readonly double with untyped property and fails


vuryss:
When creating a double of a readonly class I ger the following error:

`PHP Fatal error:  Readonly property Double\Path\To\ClassName\P1::$objectProphecyClosure must have type in /var/www/html/vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(49) : eval()'d code on line 3`

This comes from `eval` of the generated code likes this:

```
namespace Double\App\Infrastructure\Service\HttpClient\DtHttpClient {
    readonly class P1 extends \Path\To\ClassName implements \Prophecy\Prophecy\ProphecySubjectInterface, \Prophecy\Doubler\Generator\ReflectionInterface {
        private $objectProphecyClosure;

        public  function __construct(...
```

Readonly classes make all class properties readonly. And readonly properties must have a type. So this generated property `objectProphecyClosure` fails that.

It's generated in `ProphecySubjectPatch::apply`

May be the double class doesn't need to be readonly or we can make the generated fields typed?

ciaranmcnulty:
I'll be honest with you, I don't really remember how `readonly` and inheritance work together (and there's some `readonly` classes coming that'll be tricky to deal with)

If we can make the doubled properties be readonly too that's a good solution, otherwise prophecy should fail early and refuse to double the class with an appropriate message

The latter would be a good stopgap if nobody has time to look at this in detail

vuryss:
You can use an object instead to hold stuff that has to be initialized and modified later. Even if the property referencing the object is readonly - it's own data can be mutable.

BladeMF:
Same here. An argument that is set to `Argument::exact(null)` where the expected type is `?Token` is producing a double inheriting `Token`, which I don't understand why is necessary.

NicolasJourdan:
Any news about this problem ? 

@vuryss How can you use an object in order to mock the readonly class ?

WalterWoshid:
Bump, need this, too.

WalterWoshid:
For anyone needing a temporary solution:

Use it like this:

```php
class MyTest
{
    use ProphecyTrait;
    use ProphecyTraitReadonlyHack {
        ProphecyTraitReadonlyHack::prophesize insteadof ProphecyTrait;
    }

    // ...
}
```

Solution:

```php
// ProphecyTraitReadonlyHack.php

use MyApp\Prophecy\Prophet;
[...]

PR #623, opened 2024-04-25
Add Readonly support

WalterWoshid:
- closes https://github.com/phpspec/prophecy/issues/586

### DIFF
--- a/src/Prophecy/Doubler/ClassPatch/ProphecySubjectPatch.php
+++ b/src/Prophecy/Doubler/ClassPatch/ProphecySubjectPatch.php
@@ -11,11 +11,11 @@
 
 namespace Prophecy\Doubler\ClassPatch;
 
+use Prophecy\Doubler\Generator\Node\ArgumentNode;
 use Prophecy\Doubler\Generator\Node\ArgumentTypeNode;
 use Prophecy\Doubler\Generator\Node\ClassNode;
 use Prophecy\Doubler\Generator\Node\MethodNode;
-use Prophecy\Doubler\Generator\Node\ArgumentNode;
-use Prophecy\Doubler\Generator\Node\ReturnTypeNode;
+use Prophecy\Doubler\Generator\Node\PropertyTypeNode;
 
 /**
  * Add Prophecy functionality to the double.
@@ -45,10 +45,19 @@ class ProphecySubjectPatch implements ClassPatchInterface
     public function apply(ClassNode $node)
     {
         $node->addInterface('Prophecy\Prophecy\ProphecySubjectInterface');
-        $node->addProperty('objectProphecyClosure', 'private');
+        $node->addProperty(
+            'objectProphecyClosureContainer',
+            'private',
+            new PropertyTypeNode('Prophecy\Doubler\ClassPatch\ProphecySubjectPatch\ObjectProphecyClosureContainer')
+        );
 
         foreach ($node->getMethods() as $name => $method) {
             if ('__construct' === strtolower($name)) {
+                $method->setCode(
+                    $method->getCode() .
+                    '$this->objectProphecyClosureContainer = new \Prophecy\Doubler\ClassPatch\ProphecySubjectPatch\ObjectProphecyClosureContainer();'
+                );
+
                 continue;
             }
 
@@ -68,8 +77,8 @@ class ProphecySubjectPatch implements ClassPatchInterface
         $prophecyArgument->setTypeNode(new ArgumentTypeNode('Prophecy\Prophecy\ProphecyInterface'));
         $prophecySetter->addArgument($prophecyArgument);
         $prophecySetter->setCode(<<<PHP
-if (null === \$this->objectProphecyClosure) {
-    \$this->objectProphecyClosure = static function () use (\$prophecy) {
+if (null === \$this->objectProphecyClosureContainer->closure) {
+    \$this->objectProphecyClosureContainer->closure = static function () use (\$prophecy) {
         return \$prophecy;
     };
 }
@@ -77,7 +86,7 @@ PHP
     );
 
         $prophecyGetter = new MethodNode('getProphecy');
-        $prophecyGetter->setCode('return \call_user_func($this->objectProphecyClosure);');
+        $prophecyGetter->setCode('return \call_user_func($this->objectProphecyClosureContainer->closure);');
 
         if ($node->hasMethod('__call')) {
             $__call = $node->getMethod('__call');
--- a/src/Prophecy/Doubler/ClassPatch/ProphecySubjectPatch/ObjectProphecyClosureContainer.php
+++ b/src/Prophecy/Doubler/ClassPatch/ProphecySubjectPatch/ObjectProphecyClosureContainer.php
@@ -1,0 +1,27 @@
+<?php
+
+/*
+ * This file is part of the Prophecy.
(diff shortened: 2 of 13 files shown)


WalterWoshid:
Ran the tests on PHP 8.3 and PHP 7.2, both ran successfully.

src/Prophecy/Doubler/ClassPatch/ProphecySubjectPatch/ObjectProphecyClosureContainer.php
```diff
@@ -0,0 +1,22 @@
...
+/*
+ * This file is part of the Prophecy.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *     Marcello Duarte <marcello.duarte@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Prophecy\Doubler\ClassPatch\ProphecySubjectPatch;
```
stof, line 12: If this class is part of the final API, it should not be in the ClassPatch namespace at all IMO.

And this class should probably be tagged as `@internal`

src/Prophecy/Doubler/Generator/Node/ClassNode.php
```diff
@@ -103,34 +101,21 @@ public function hasInterface($interface)
     }
 
     /**
-     * @return array<string, string>
-     *
-     * @phpstan-return array<string, 'public'|'private'|'protected'>
+     * @return array<string, PropertyNode>
```
stof, line 104: This cannot be done as it. This ClassNode is not internal (it is part of the API exposed to custom class patches) so we must preserve backward compatibility. Changing the return type is a BC break.

src/Prophecy/Doubler/Generator/Node/ClassNode.php
```diff
@@ -103,34 +101,21 @@ public function hasInterface($interface)
...
-     * @param string $name
-     * @param string $visibility
+     * @param PropertyNode $property
      *
      * @return void
-     *
-     * @phpstan-param 'public'|'private'|'protected' $visibility
      */
-    public function addProperty($name, $visibility = 'public')
+    public function addProperty(PropertyNode $property): void
```
stof, line 116: This cannot be done as it. This ClassNode is not internal (it is part of the API exposed to custom class patches) so we must preserve backward compatibility. Changing the argument types is a BC break.

For reference, the Symfony team made some documentation about which kind of changes are allowed when working on the code to avoid breaking backward compatibility: https://symfony.com/doc/current/contributing/code/bc.html#working-on-symfony-code (note that those answers assume that we want to provide the guarantees [listed above](https://symfony.com/doc/current/contributing/code/bc.html#using-symfony-code). If you want to turn the few remaining "no" in that second to also guarantee BC on them, it would have a big impact on the changes allowed)

stof (changes requested):

src/Prophecy/Doubler/ClassPatch/ProphecySubjectPatch/ObjectProphecyClosureContainer.php
WalterWoshid: What defines the final API? Where should I place it instead?

src/Prophecy/Doubler/Generator/Node/ClassNode.php
WalterWoshid: No offense, but I think Symfonys rules are stupid, without breaking backward compatibility, code will look ugly.

stof: This is not about making it ugly or not ugly. It is about not breaking projects using our API.

Note that the documentation on the Symfony website is not based on "we want to do that". It is based on "we want to guarantee that level of backward compatibility for users of our code". And backward compatibility is about allowing projects to update their dependencies without having to rewrite their code.
Each time you use a `^` operator in your composer constraints, you rely on packages following semantic versioning. And backward compatibility is at the heart of semver (unless you decide to only release major version all the time, but then downstream projects will see no different between using a semver constraint or an exact match constraint).

WalterWoshid: I fully agree on that and it's probably the right way for packages, I just don't like it.

src/Prophecy/Doubler/Generator/Node/ClassNode.php
WalterWoshid: How about now?

src/Prophecy/Doubler/ClassPatch/ProphecySubjectPatch.php
```diff
@@ -45,10 +45,19 @@ public function supports(ClassNode $node)
...
+            'objectProphecyClosureContainer',
+            'private',
+            new PropertyTypeNode('Prophecy\Doubler\ClassPatch\ProphecySubjectPatch\ObjectProphecyClosureContainer')
+        );
 
         foreach ($node->getMethods() as $name => $method) {
             if ('__construct' === strtolower($name)) {
+                $method->setCode(
+                    $method->getCode() .
+                    '$this->objectProphecyClosureContainer = new \Prophecy\Doubler\ClassPatch\ProphecySubjectPatch\ObjectProphecyClosureContainer();'
```
mathroc, line 58: should this be at the beginning in case there's an early return?

src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
```diff
@@ -50,8 +49,8 @@ public function generate($classname, Node\ClassNode $class)
             )
         );
 
-        foreach ($class->getProperties() as $name => $visibility) {
-            $code .= sprintf("%s \$%s;\n", $visibility, $name);
+        foreach ((array) $class->getPropertyNodes() as $propertyNode) {
```
Nek-, line 52: `getPropertyNodes()` returns an array anyway, why casting it?

src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
```diff
@@ -50,8 +49,8 @@ public function generate($classname, Node\ClassNode $class)
...
+            $code .= $this->generateProperty($propertyNode)."\n";
```
Nek-, line 53: For new Type nodes, we use the `__toString()` method. I believe that keeping the same generation way between classes impacts the maintainability so it would be better to use a toString() method here as well.

src/Prophecy/Doubler/Generator/Node/PropertyNode.php
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * @param string $visibility
+     *
+     * @return void
+     *
+     * @phpstan-param 'public'|'private'|'protected' $visibility
+     */
+    public function setVisibility(string $visibility)
+    {
+        $visibility = strtolower($visibility);
```
Nek-, line 87: why strtolower here?

src/Prophecy/Doubler/Generator/Node/ClassNode.php
Nek-: It's not possible to break the backward compatibility mostly because of the cross dependency between phpspec et prophecy. But it's also best to not break for user experience as well. (we didn't break for #637 , I don't think we will break here either!)

Nek-:
@WalterWoshid is it fine to you if I take your work to make a PR that will be merged?

Issue #624, opened 2024-04-30
PHP 8.4 support


andypost:
I started testing Drupal 11 with PHP 8.4 and faced issues because of https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated

I did patch some code (will provide PR) but still getting deprecation warnings from mocks

Probably the `\Prophecy\Doubler\ClassPatch\DisableConstructorPatch::apply()` needs changes

```
    OK (6 tests, 154 assertions)
    
    Other deprecation notices (2)
    
      1x:
    Double\Drupal\Core\Cache\Context\CacheContextsManager\P1::__construct():
    Implicitly marking parameter $container as nullable is deprecated, the
    explicit nullable type must be used instead
        1x in VariationCacheTest::setUp from Drupal\Tests\Core\Cache
    
      1x:
    Double\Drupal\Core\Cache\Context\CacheContextsManager\P1::__construct():
    Implicitly marking parameter $contexts as nullable is deprecated, the
    explicit nullable type must be used instead
        1x in VariationCacheTest::setUp from Drupal\Tests\Core\Cache
```

andypost:
Now waiting for new release but looks it depends on compatible `phpspec/phpspec` release https://github.com/phpspec/phpspec/pull/1472#issuecomment-2444216227

andypost:
@stof last 8.4.0RC4 this week any chance to see compatible release next week? This is only Drupal dependency which reports incompatibility with 8.4

andypost:
@stof Thank you a lot!!!

PR #625, opened 2024-04-30
 Fix deprecated nullable areguments #624 

andypost:
Related to #624

### DIFF
--- a/spec/Prophecy/Doubler/Generator/ClassCodeGeneratorSpec.php
+++ b/spec/Prophecy/Doubler/Generator/ClassCodeGeneratorSpec.php
@@ -75,8 +75,7 @@ class ClassCodeGeneratorSpec extends ObjectBehavior
         $method5->getCode()->willReturn('return;');
 
         $argument11->getName()->willReturn('fullname');
-        $argument11->isOptional()->willReturn(true);
-        $argument11->getDefault()->willReturn(null);
+        $argument11->isOptional()->willReturn(false);
         $argument11->isPassedByReference()->willReturn(false);
         $argument11->isVariadic()->willReturn(false);
         $argument11->getTypeNode()->willReturn(new ArgumentTypeNode('array'));
@@ -116,7 +115,7 @@ class CustomClass extends \RuntimeException implements \Prophecy\Doubler\Generat
 public $name;
 private $email;
 
-public static function getName(array $fullname = NULL, \ReflectionClass $class, object $instance): ?string {
+public static function getName(array $fullname, \ReflectionClass $class, object $instance): ?string {
 return $this->name;
 }
 protected  function getEmail(?string $default = 'ever.zet@gmail.com') {
@@ -265,14 +264,14 @@ PHP;
         $argument->getDefault()->willReturn(null);
         $argument->isPassedByReference()->willReturn(true);
         $argument->isVariadic()->willReturn(false);
-        $argument->getTypeNode()->willReturn(new ArgumentTypeNode('array'));
+        $argument->getTypeNode()->willReturn(new ArgumentTypeNode('array', 'null'));
 
         $code = $this->generate('CustomClass', $class);
         $expected = <<<'PHP'
 namespace  {
 class CustomClass extends \RuntimeException implements \Prophecy\Doubler\Generator\MirroredInterface {
 
-public  function getName(array &$fullname = NULL) {
+public  function getName(?array &$fullname = NULL) {
 return $this->name;
 }
 


src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
```diff
@@ -101,7 +101,9 @@ private function generateArguments(array $arguments): array
     {
         return array_map(function (Node\ArgumentNode $argument){
 
-            $php = $this->generateTypes($argument->getTypeNode());
+            $types = $argument->getTypeNode()->getTypes();
+            $null = $argument->isOptional() && $argument->getDefault() === NULL && \count($types) === 1 && $types[0] !== 'mixed';
+            $php = $this->generateTypes($argument->getTypeNode(), $null);
```
andypost, line 106: it needs polishing

jrfnl:
Would be lovely if this PR could get finished off and merged (and released!) as anyone who is currently using PHPUnit 9 to assess their own project's PHP 8.4 readiness is faced with a wall of deprecation notices coming from Prophecy.

@andypost Anything you need help with ? Want me to review something ?

andypost:
@jrfnl thank, please help me to improve the code as I don't like amount of `\count()` added to separate fixes for constructors and return values

andypost:
Personally I do use it as a part of patch for Drupal compatibility [issue](https://www.drupal.org/project/drupal/issues/3427903)

stof:
To make things easier to review, I'd like to see this PR split in 2:

- one changing the parameter type in the prophecy code itself (which can be the result of running php-cs-fixer's dedicated rule on the codebase)
- one changing the logic for generated code.

andypost:
I did split code-style related fixes into https://github.com/phpspec/prophecy/pull/630

stof:
Please rebase this branch now that the other PR doing the CS changes on the codebase is merged.

stof:
@andypost what is the case which requires the change in this PR ? Is it for any nullable optional argument, or only for cases where the original code is relying on an implicit nullable type ?

andypost:
@stof example failures are 

The test case [code](https://git.drupalcode.org/project/drupal/-/blob/e5653c11cd20e66831aca7e7a7568717baff5d9d/core/tests/Drupal/KernelTests/Core/Bootstrap/ExtensionPathResolverTest.php#L99-109)
```php
  public function testUnknownExtension(): void {
    $module_extension_list = $this->prophesize(ModuleExtensionList::class);
    $profile_extension_list = $this->prophesize(ProfileExtensionList::class);
    $theme_extension_list = $this->prophesize(ThemeExtensionList::class);
    $theme_engine_extension_list = $this->prophesize(ThemeEngineExtensionList::class);
    $resolver = new ExtensionPathResolver(
      $module_extension_list->reveal(),
      $profile_extension_list->reveal(),
      $theme_extension_list->reveal(),
      $theme_engine_extension_list->reveal()
    );

```

produce [output](https://git.drupalcode.org/issue/drupal-3427903/-/jobs/2281897#L3112) 
```
    /builds/issue/drupal-3427903/vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(44)
    : eval()'d code:5
    Double\Drupal\Core\Extension\ModuleExtensionList\P1::__construct():
    Implicitly marking parameter $cache as nullable is deprecated, the explicit
    nullable type must be used instead
    
    Triggered by:
    
    *
    Drupal\KernelTests\Core\Bootstrap\ExtensionPathResolverTest::testUnknownExtension
     
    /builds/issue/drupal-3427903/core/tests/Drupal/KernelTests/Core/Bootstrap/ExtensionPathResolverTest.php:99
```

andypost:
As I get it... when class with required typed constructor argument is mocked, prophecy making argument nullable but some how without leading `?`

stof:
@andypost can you show the class definition of `ModuleExtensionList` ?

andypost:
Even better link https://github.com/drupal/drupal/blob/e5653c11cd20e66831aca7e7a7568717baff5d9d/core/lib/Drupal/Core/Extension/ModuleExtensionList.php#L20

andypost:
Fixed nullable types in spec and now it pass

stof:
@andypost if you update the `$container_modules_info` argument of the `ModuleExtensionList` constructor to avoid using an implicit nullable type, does the double still use an implicit nullable type (without this PR applied) ?

andypost:
@stof The `$container_modules_info` supposed to have default value and is not nullable and without patch `reveal()` just setting to `null` the `CacheBackendInterface $cache` but so my patch adding `?` for such cases

andypost:
The error is the same
```
    Double\Drupal\Core\Extension\ModuleExtensionList\P1::__construct():
    Implicitly marking parameter $cache as nullable is deprecated, the explicit
    nullable type must be used instead
```

src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
```diff
@@ -78,14 +79,14 @@ private function generateMethod(Node\MethodNode $method): string
...
-    private function generateTypes(TypeNodeAbstract $typeNode): string
+    private function generateTypes(TypeNodeAbstract $typeNode, bool $nullable = FALSE): string
     {
         if (!$typeNode->getTypes()) {
             return '';
         }
 
         // When we require PHP 8 we can stop generating ?foo nullables and remove this first block
-        if ($typeNode->canUseNullShorthand()) {
+        if ($typeNode->canUseNullShorthand() || $nullable) {
```
andypost, line 89: this place exactly is not working, it just making all arguments nullable (optional) but if the type defined for one then no `?` is added

andypost: Specifically for case when type is `mixed` and no `null` (PHP 8.2) defined in intersection (PHP 8.1)

src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
```diff
@@ -101,7 +102,18 @@ private function generateArguments(array $arguments): array
...
+              $count = \count($types);
+              if ($count === 1 && $types[0] === 'mixed' ) {
+                $nullable = FALSE;
+              }
+              elseif ($count > 1 && !isset($types['null'])) {
+                $argument->setTypeNode(new ArgumentTypeNode('null', ...$argument->getTypeNode()->getNonNullTypes()));
+                $nullable = FALSE;
+              }
+            }
+            $php = $this->generateTypes($argument->getTypeNode(), $nullable);
```
andypost, line 116: new argument required because `$argument->getTypeNode()` has no access to default value and optional sproperty

stof:
OK, apparently, those errors come from the DisableConstructorPatch which aims at making all constructor arguments optional.

the right fix is to update the implementation of that class patch so that it makes all arguments nullable as well (by updating the argument type) instead of adding a `null` default on non-nullable types. This would be much cleaner than making the code generator turn implicit nullable types to explicit ones.

andypost:
@stof I can't make such big refactoring and need help to cover all cases

Jean85 (approved):
This + #632 should solve the PHP 8.4 compat

stof:
@andypost now that the DisableConstructorPatch has been updated in #632, is there anything left in that PR that should be merged ?

Jean85:
This PR seems to fix the same issue (implicit nullable arguments deprecation) but with generated methods, not constructors, so I think it should be merged.

The problem reported in https://github.com/phpspec/prophecy/pull/625#issuecomment-2317858043 was covered by #632.

andypost:
I bet both issues solving compatibility as I have no 8.4 code yet (property getting unexplored yet)
So looking for minor release or kind of it, also looking for feedback as next RC just arrived!

spec/Prophecy/Doubler/Generator/ClassCodeGeneratorSpec.php
```diff
@@ -272,7 +272,7 @@ function it_overrides_properly_methods_with_args_passed_by_reference(
 namespace  {
 class CustomClass extends \RuntimeException implements \Prophecy\Doubler\Generator\MirroredInterface {
 
-public  function getName(array &$fullname = NULL) {
+public  function getName(?array &$fullname = NULL) {
```
andypost, line 275: @stof this changes still required to prevent deprecations

stof: Can you update the PR to keep only the required changes ?

andypost: updated! looking for unblocking release)

spec/Prophecy/Doubler/Generator/ClassCodeGeneratorSpec.php
```diff
@@ -116,7 +116,7 @@ class CustomClass extends \RuntimeException implements \Prophecy\Doubler\Generat
 public $name;
 private $email;
 
-public static function getName(array $fullname = NULL, \ReflectionClass $class, object $instance): ?string {
+public static function getName(?array $fullname = NULL, \ReflectionClass $class, object $instance): ?string {
```
andypost, line 119: @stof it looks weird that optional argument going before other ones, maybe redo it?

stof: I suggest removing the default value entirely in this `fullname` argument, making it a nullable required argument (which is what it was in practice)

stof (approved):

PR #630, opened 2024-08-21
PHPCS: nullable types PHP 8.4

andypost:
First part for #624

### DIFF
--- a/fixtures/WithArguments.php
+++ b/fixtures/WithArguments.php
@@ -4,7 +4,7 @@ namespace Fixtures\Prophecy;
 
 class WithArguments
 {
-    public function methodWithArgs(\ArrayAccess $arg_1, array $arg_2 = [], \ArrayAccess $arg_3 = null)
+    public function methodWithArgs(\ArrayAccess $arg_1, array $arg_2 = [], ?\ArrayAccess $arg_3 = null)
     {
     }
     
--- a/fixtures/WithCallableArgument.php
+++ b/fixtures/WithCallableArgument.php
@@ -4,7 +4,7 @@ namespace Fixtures\Prophecy;
 
 class WithCallableArgument
 {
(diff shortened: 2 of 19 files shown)


.github/workflows/build.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: ["7.2", "7.3", "7.4", "8.0", "8.1", "8.2", "8.3"]
+        php: ["7.2", "7.3", "7.4", "8.0", "8.1", "8.2", "8.3", "8.4"]
```
stof, line 17: This should not be done in this PR applying a CS change (especially as it does not work as is). Keep this PR focused only on the CS change.

stof (commented):
This PR should also fix the new deprecations reported by a optional argument before required ones (which was skipped when the default value was making the type nullable implicitly).

Note that PHP-CS-Fixer has a (risky) rule `no_unreachable_default_argument_value` for that as well

.github/workflows/build.yml
andypost: removed the change, supposed to try run pipelines

stof: I opened https://github.com/phpspec/prophecy/pull/631 to add new CI jobs

andypost: Rebased on top of it

jrfnl:
> This PR should also fix the new deprecations reported by a optional argument before required ones (which was skipped when the default value was making the type nullable implicitly).

I scanned this branch with PHPCompatibility, which can check for both deprecations and this branch is clean of both of those issues.

stof:
> Edit: at least it is at this time (when I scanned it).

It was not the case when I asked for that change, as shown by the static analysis CI job at that time: https://github.com/phpspec/prophecy/actions/runs/10492742258/job/29065879468

stof (approved):

PR #632, opened 2024-10-03
DisableConstructorPatch: make optional constructor parameters nullable explicitly

W0rma:
This PR tries to fix https://github.com/phpspec/prophecy/pull/625#issuecomment-2304548400


src/Prophecy/Doubler/ClassPatch/DisableConstructorPatch.php
```diff
@@ -55,6 +56,16 @@ public function apply(ClassNode $node)
         \assert($constructor !== null);
         foreach ($constructor->getArguments() as $argument) {
             $argument->setDefault(null);
+            if (\PHP_VERSION_ID < 80400) {
```
stof, line 59: I suggest doing that for all versions instead, to avoid different behavior between PHP versions (making a type explicitly nullable is supported on PHP 7.1+, which covers more than our target versions)

src/Prophecy/Doubler/ClassPatch/DisableConstructorPatch.php
```diff
@@ -55,6 +56,16 @@ public function apply(ClassNode $node)
...
+                continue;
+            }
+
+            $types = $argument->getTypeNode()->getNonNullTypes();
+            if ([] == $types || ['mixed'] == $types) {
```
stof, line 64: please use strict comparisons here

src/Prophecy/Doubler/ClassPatch/DisableConstructorPatch.php
W0rma: fixed

W0rma:
https://github.com/phpspec/prophecy/pull/633 should fix the CI (cs fixer + phpstan)

Jean85 (approved):

Jean85:
This should receive the PHP 8.4 label

Issue #558, opened 2022-08-22
Support for Disjunctive Normal Form Types


stof:
See https://wiki.php.net/rfc/dnf_types

We need to make sure they can be doubled properly, with tests ensuring they work fine.

kschatzle:
Noting that getTypes() for the following code may now (PHP8.2) return 'ReflectionIntersectionType':

https://github.com/phpspec/prophecy/blob/a4477cd17b481a6095494c9583316a3e3aa88916/src/Prophecy/Prophecy/MethodProphecy.php#L90-L92

https://github.com/phpspec/prophecy/blob/a4477cd17b481a6095494c9583316a3e3aa88916/src/Prophecy/Doubler/Generator/ClassMirror.php#L224-L226

I'm unsure what the underlying decision to not support `ReflectionIntersectionType` was. My assumption is that it was difficult (maybe impossible) to create the double with the necessary values.

Possible solution would be to loop through types looking for `ReflectionIntersectionType` and throwing a nice Exception.

It does throw an err.  The class effectively turns this:
`
    public function doSomething () : A|(B&D)|null
    {

    }
`
Into
`
        public  function doSomething(): A|B&D|null { //At least one err here.
            return $this->getProphecy()->makeProphecyMethodCall(__FUNCTION__, func_get_args());
        }
`

All this said, I'm confused by something, so going to ping the php-src group. Either found a bug or something I don't understand.

ciaranmcnulty:
> I'm unsure what the underlying decision to not support ReflectionIntersectionType was

Simply that PHP 8.1 was coming and we wanted a clear exception to be thrown here rather than a hard error


Nobody's made it work since (or TBH has raised an issue)

ciaranmcnulty:
if someone wants to make intersection types work that's fine, to be clear

stof:
The issue tracking proper support for intersection types is #535, but nobody worked on it yet.

PR #637, opened 2025-04-17
Intersection types

### LINKED ISSUE #535: Intersection types in generated code

Nek-:
Note about the design: I created a new sub-api for type management because the current one is not compatible with intersections but needs to exist to keep backward compatibility.

What still needs to be done:
- [x] Fix existing bugs/tests with this implementation (god it's hard to fix)
- [x] Add deprecation about the old API (some comments already existing in the code will help!)
- [x] Refactor to remove deprecated calls
- [x] Implement code generation with intersection
- [x] Add tests regarding intersection
- [x] Implement intersection support & dnf for methods args

Please have a look and tell me if this kind of implementation would be ok for you. Thanks!

Stands as replacement for #569 and should fix #535 and #558

-------------------------

Update: the PR is ready and waiting to be merged :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,8 @@
         "phpdocumentor/reflection-docblock": "^5.2",
         "sebastian/comparator":              "^3.0 || ^4.0 || ^5.0 || ^6.0 || ^7.0",
         "doctrine/instantiator":             "^1.2 || ^2.0",
-        "sebastian/recursion-context":       "^3.0 || ^4.0 || ^5.0 || ^6.0 || ^7.0"
+        "sebastian/recursion-context":       "^3.0 || ^4.0 || ^5.0 || ^6.0 || ^7.0",
+        "symfony/deprecation-contracts":     "^2.5 || ^3.1"
     },
 
     "require-dev": {
--- a/spec/Prophecy/Doubler/Generator/ClassCodeGeneratorSpec.php
+++ b/spec/Prophecy/Doubler/Generator/ClassCodeGeneratorSpec.php
@@ -10,6 +10,11 @@ use Prophecy\Doubler\Generator\Node\ArgumentTypeNode;
 use Prophecy\Doubler\Generator\Node\ClassNode;
 use Prophecy\Doubler\Generator\Node\MethodNode;
 use Prophecy\Doubler\Generator\Node\ReturnTypeNode;
+use Prophecy\Doubler\Generator\Node\Type\BuiltinType;
+use Prophecy\Doubler\Generator\Node\Type\IntersectionType;
+use Prophecy\Doubler\Generator\Node\Type\ObjectType;
+use Prophecy\Doubler\Generator\Node\Type\SimpleType;
+use Prophecy\Doubler\Generator\Node\Type\UnionType;
 
 class ClassCodeGeneratorSpec extends ObjectBehavior
 {
@@ -39,7 +44,10 @@ class ClassCodeGeneratorSpec extends ObjectBehavior
         $method1->returnsReference()->willReturn(false);
         $method1->isStatic()->willReturn(true);
         $method1->getArguments()->willReturn(array($argument11, $argument12, $argument13));
-        $method1->getReturnTypeNode()->willReturn(new ReturnTypeNode('string', 'null'));
+        $method1->getReturnTypeNode()->willReturn(new ReturnTypeNode(new UnionType([
+            new BuiltinType('string'),
+            new BuiltinType('null'),
+        ])));
         $method1->getCode()->willReturn('return $this->name;');
 
         $method2->getName()->willReturn('getEmail');
@@ -55,7 +63,7 @@ class ClassCodeGeneratorSpec extends ObjectBehavior
         $method3->returnsReference()->willReturn(true);
         $method3->isStatic()->willReturn(false);
         $method3->getArguments()->willReturn(array($argument31));
-        $method3->getReturnTypeNode()->willReturn(new ReturnTypeNode('string'));
+        $method3->getReturnTypeNode()->willReturn(new ReturnTypeNode(new BuiltinType('string')));
         $method3->getCode()->willReturn('return $this->refValue;');
 
         $method4->getName()->willReturn('doSomething');
@@ -63,7 +71,7 @@ class ClassCodeGeneratorSpec extends ObjectBehavior
         $method4->returnsReference()->willReturn(false);
         $method4->isStatic()->willReturn(false);
         $method4->getArguments()->willReturn(array());
-        $method4->getReturnTypeNode()->willReturn(new ReturnTypeNode('void'));
+        $method4->getReturnTypeNode()->willReturn(new ReturnTypeNode(new BuiltinType('void')));
         $method4->getCode()->willReturn('return;');
 
         $method5->getName()->willReturn('returnObject');
@@ -71,33 +79,36 @@ class ClassCodeGeneratorSpec extends ObjectBehavior
         $method5->returnsReference()->willReturn(false);
         $method5->isStatic()->willReturn(false);
         $method5->getArguments()->willReturn(array());
-        $method5->getReturnTypeNode()->willReturn(new ReturnTypeNode('object'));
+        $method5->getReturnTypeNode()->willReturn(new ReturnTypeNode(new BuiltinType('object')));
         $method5->getCode()->willReturn('return;');
 
         $argument11->getName()->willReturn('fullname');
         $argument11->isOptional()->willReturn(false);
         $argument11->isPassedByReference()->willReturn(false);
         $argument11->isVariadic()->willReturn(false);
-        $argument11->getTypeNode()->willReturn(new ArgumentTypeNode('array'));
+        $argument11->getTypeNode()->willReturn(new ArgumentTypeNode(new BuiltinType('array')));
 
         $argument12->getName()->willReturn('class');
         $argument12->isOptional()->willReturn(false);
         $argument12->isPassedByReference()->willReturn(false);
         $argument12->isVariadic()->willReturn(false);
-        $argument12->getTypeNode()->willReturn(new ArgumentTypeNode('ReflectionClass'));
+        $argument12->getTypeNode()->willReturn(new ArgumentTypeNode(new ObjectType('ReflectionClass')));
 
         $argument13->getName()->willReturn('instance');
         $argument13->isOptional()->willReturn(false);
         $argument13->isPassedByReference()->willReturn(false);
         $argument13->isVariadic()->willReturn(false);
-        $argument13->getTypeNode()->willReturn(new ArgumentTypeNode('object'));
+        $argument13->getTypeNode()->willReturn(new ArgumentTypeNode(new BuiltinType('object')));
 
         $argument21->getName()->willReturn('default');
         $argument21->isOptional()->willReturn(true);
         $argument21->getDefault()->willReturn('ever.zet@gmail.com');
         $argument21->isPassedByReference()->willReturn(false);
         $argument21->isVariadic()->willReturn(false);
-        $argument21->getTypeNode()->willReturn(new ArgumentTypeNode('string', 'null'));
+        $argument21->getTypeNode()->willReturn(new ArgumentTypeNode(new UnionType([
+            new BuiltinType('string'),
+            new BuiltinType('null'),
+        ])));
 
         $argument31->getName()->willReturn('refValue');
         $argument31->isOptional()->willReturn(false);
@@ -115,10 +126,10 @@ class CustomClass extends \RuntimeException implements \Prophecy\Doubler\Generat
 public $name;
 private $email;
 
-public static function getName(array $fullname, \ReflectionClass $class, object $instance): ?string {
+public static function getName(array $fullname, \ReflectionClass $class, object $instance): string|null {
 return $this->name;
 }
-protected  function getEmail(?string $default = 'ever.zet@gmail.com') {
+protected  function getEmail(string|null $default = 'ever.zet@gmail.com') {
 return $this->email;
 }
 public  function &getRefValue( $refValue): string {
@@ -206,13 +217,13 @@ PHP;
         $argument3->isOptional()->willReturn(false);
         $argument3->isPassedByReference()->willReturn(false);
         $argument3->isVariadic()->willReturn(true);
-        $argument3->getTypeNode()->willReturn(new ArgumentTypeNode('ReflectionClass'));
+        $argument3->getTypeNode()->willReturn(new ArgumentTypeNode(new ObjectType('ReflectionClass')));
 
         $argument4->getName()->willReturn('args');
         $argument4->isOptional()->willReturn(false);
         $argument4->isPassedByReference()->willReturn(true);
         $argument4->isVariadic()->willReturn(true);
-        $argument4->getTypeNode()->willReturn(new ArgumentTypeNode('ReflectionClass'));
+        $argument4->getTypeNode()->willReturn(new ArgumentTypeNode(new ObjectType('ReflectionClass')));
 
 
         $code = $this->generate('CustomClass', $class);
@@ -264,14 +275,17 @@ PHP;
         $argument->getDefault()->willReturn(null);
         $argument->isPassedByReference()->willReturn(true);
         $argument->isVariadic()->willReturn(false);
-        $argument->getTypeNode()->willReturn(new ArgumentTypeNode('array', 'null'));
+        $argument->getTypeNode()->willReturn(new ArgumentTypeNode(new UnionType([
+            new BuiltinType('array'),
+            new BuiltinType('null'),
+        ])));
 
         $code = $this->generate('CustomClass', $class);
         $expected = <<<'PHP'
 namespace  {
 class CustomClass extends \RuntimeException implements \Prophecy\Doubler\Generator\MirroredInterface {
 
-public  function getName(?array &$fullname = NULL) {
+public  function getName(array|null &$fullname = NULL) {
 return $this->name;
 }
 
@@ -296,7 +310,11 @@ PHP;
         $method->getVisibility()->willReturn('public');
         $method->isStatic()->willReturn(false);
         $method->getArguments()->willReturn([]);
-        $method->getReturnTypeNode()->willReturn(new ReturnTypeNode('int', 'string', 'null'));
+        $method->getReturnTypeNode()->willReturn(new ReturnTypeNode(new UnionType([
+            new BuiltinType('int'),
+            new BuiltinType('string'),
+            new BuiltinType('null'),
+        ])));
         $method->returnsReference()->willReturn(false);
         $method->getCode()->willReturn('');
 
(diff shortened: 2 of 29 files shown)


src/Prophecy/Doubler/Generator/ClassMirror.php
```diff
@@ -158,14 +162,22 @@ private function reflectMethodToNode(ReflectionMethod $method, Node\ClassNode $c
...
-            $node->setReturnTypeNode(new ReturnTypeNode(...$returnTypes));
+            $returnReflectionType = $method->getReturnType();
+            \assert($returnReflectionType !== null);
         } elseif (method_exists($method, 'hasTentativeReturnType') && $method->hasTentativeReturnType()) {
-            \assert($method->getTentativeReturnType() !== null);
-            $returnTypes = $this->getTypeHints($method->getTentativeReturnType(), $method->getDeclaringClass(), $method->getTentativeReturnType()->allowsNull());
-            $node->setReturnTypeNode(new ReturnTypeNode(...$returnTypes));
+            // Tentative return types also need reflection
+            $returnReflectionType = $method->getTentativeReturnType();
+            \assert($returnReflectionType !== null);
```
stof, line 172: I think those asserts are useless (the `$returnReflectionType` variable will still be nullable anyway as there is no `else` branch assigning it)

src/Prophecy/Doubler/Generator/ClassMirror.php
```diff
@@ -232,6 +243,74 @@ private function getDefaultValue(ReflectionParameter $parameter)
...
     }
 
+    /**
+     * @param ReflectionClass<object> $declaringClass Context reflection class
+     */
+    private function createTypeFromReflection(ReflectionType $type, ReflectionClass $declaringClass): TypeInterface
+    {
+        if ($type instanceof ReflectionIntersectionType) {
+            foreach ($type->getTypes() as $innerReflectionType) {
+                $innerTypes[] = new SimpleType($innerReflectionType->getName());
```
stof, line 253: isn't this missing the resolution of the type name ?

stof: maybe this should use `createTypeFromReflection` for the inner type as well.

src/Prophecy/Doubler/Generator/Node/TypeNodeAbstract.php
```diff
@@ -2,45 +2,124 @@
...
 {
-    /** @var array<string, string> */
-    protected $types = [];
+    protected TypeInterface|null $type;
 
-    public function __construct(string ...$types)
+    /**
+     * @param string|TypeInterface ...$types
+     */
+    public function __construct(string|TypeInterface ...$types)
```
stof, line 18: shouldn't the new API allows `null` as well ?

composer.json
```diff
@@ -18,7 +18,7 @@
     ],
 
     "require": {
-        "php":                               "^7.2 || 8.0.* || 8.1.* || 8.2.* || 8.3.* || 8.4.*",
+        "php":                               "8.2.* || 8.3.* || 8.4.*",
```
stof, line 21: Changing the supported PHP versions should likely be done in a dedicated PR, to avoid hiding such impactful change (and to properly clean things that are not needed anymore, like `method_exists($method, 'hasTentativeReturnType')` for instance)

src/Prophecy/Doubler/Generator/Node/TypeNodeAbstract.php
```diff
@@ -74,44 +153,42 @@ protected function getRealType(string $type): string
...
+                // Class / Interface type
                 return $this->prefixWithNsSeparator($type);
         }
     }
 
     /**
+     * @todo: put this in SimpleType
      * @return void
      */
     protected function guardIsValidType()
```
stof, line 170: Be careful. ReturnTypeNode extends this class to override this method, but you still use `$this->types` in it.

src/Prophecy/Doubler/Generator/Node/TypeNodeAbstract.php
Nek-: Yep the new API will be `?string|TypeInterface`, you're absolutely right

src/Prophecy/Doubler/Generator/Node/TypeNodeAbstract.php
Nek-: After cleanup this method should not even exist anymore here.

src/Prophecy/Doubler/Generator/Node/TypeNodeAbstract.php
```diff
@@ -74,44 +153,42 @@ protected function getRealType(string $type): string
...
+            case 'never':
+                return $type;
             default:
+                // Class / Interface type
                 return $this->prefixWithNsSeparator($type);
         }
     }
 
     /**
+     * @todo: put this in SimpleType
```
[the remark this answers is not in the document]
Nek-: I guess each type will have its own exceptions, this still needs to be done.

src/Prophecy/Doubler/Generator/Node/TypeNodeAbstract.php
```diff
@@ -74,44 +153,42 @@ protected function getRealType(string $type): string
...
                 throw new DoubleException('Type cannot be standalone true');
             }
 
-            if ($this->types == ['true' => 'true', 'null' => 'null']) {
+            if ($this->type->equals(new UnionType([new SimpleType('true'), new SimpleType('null')]))) {
                 throw new DoubleException('Type cannot be nullable true');
             }
         }
-
-        if (\PHP_VERSION_ID >= 80000 && isset($this->types['mixed']) && count($this->types) !== 1) {
```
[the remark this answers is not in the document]
Nek-: Question is more like why is this method `guardIsValidType` still here. I will remove it on cleanup. The UnionType checks this mixed-thing already.

Nek-:
Thank you for your review @stof . And good news! I fixed the bugs in my previous implementation. (some you noticed in your review!) Keeping backward compatibility has been a real pain! But it WORKS.

It's currently a fully working PR for intersection type.... But only for return types!

Some work needs to be done for cleaning... But it also needs php to be bumped to 8.1+ (which would probably be better in another PR), and a lot of cleaning.

But for now, since it's a lot of work and personal investment, I'd like you to tell me if the implementation is ok before working on it again.

src/Prophecy/Doubler/Generator/ClassMirror.php
```diff
@@ -232,6 +243,74 @@ private function getDefaultValue(ReflectionParameter $parameter)
...
+            $simpleType = new SimpleType($name); // SimpleType constructor normalizes
+
+            // Handle nullability for named types explicitly by wrapping in a UnionType if needed
+            if ($type->allowsNull() && $name !== 'mixed' && $name !== 'null') {
+                // Check if SimpleType already resolved to 'null' (e.g. input was 'null')
+                if ($simpleType->getType() === 'null') {
+                    return $simpleType; // Already null, no union needed
+                }
+                // Check if SimpleType already resolved to 'mixed'
+                if ($simpleType->getType() === 'mixed') {
```
stof, line 284: isn't this dead code due to the checks on `$name` on the wrapping condition ?

src/Prophecy/Doubler/Generator/Node/Type/UnionType.php
```diff
@@ -0,0 +1,96 @@
+<?php
+
+namespace Prophecy\Doubler\Generator\Node\Type;
+
+use Prophecy\Exception\Doubler\DoubleException;
+
+class UnionType implements TypeInterface
```
stof, line 7: Please also mark those classes as `final` (or at least as `@final` in phpdoc if we have an actual need to create doubles for them in our own tests). It will simplify our maintenance for the future in term of BC needs

src/Prophecy/Doubler/Generator/Node/Type/SimpleType.php
```diff
@@ -0,0 +1,84 @@
+<?php
+
+namespace Prophecy\Doubler\Generator\Node\Type;
+
+class SimpleType implements TypeInterface, \Stringable
```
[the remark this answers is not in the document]
Nek-: Having an ObjectType and BuiltinType looks over-engineered to me (a lot of code for basically the same thing). It also involve a factory in the middle making the distinction between the 2 cases (in classmirror).

About the problem you mention, it would be possible to normalize the type only in the toString method, isn't it?

I understand that having the 2 types makes sense. I actually like pretty much the idea, I just worry of the "why" and "isn't it just more code for more code".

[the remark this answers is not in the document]
stof: ClassMirror could create the right type based on `ReflectionNamedType::isBuiltIn()`, which would also mean we automatically get support for new builtin types for newer PHP versions instead of having to hardcode which strings are a built-in type that need to produce a different code representation (classes need to be prefixed with `\` to be used in a namespaced context while built-in type must not be prefixed).

src/Prophecy/Doubler/Generator/Node/Type/SimpleType.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        return $this->builtin;
+    }
+
+    private function normalizeType(string $type): string
+    {
+        $this->builtin = true;
+        switch ($type) {
+            // type aliases
+            case 'double':
```
stof, line 24: `double` is **not** a builtin alias in PHP. It is treated as a class name (and triggers a warning). Same for the others: https://3v4l.org/T9iRf

I think the new API should not have those special cases (the BC layer in TypeNodeAbstract should handle it instead, and triggering an additional deprecation saying that `BuiltinType` in the new API will _not_ support such aliases and they should use the proper value instead)

stof (changes requested):

spec/Prophecy/Doubler/Generator/ClassCodeGeneratorSpec.php
```diff
@@ -310,6 +313,47 @@ public  function foo(): int|string|null {
...
+        $method->returnsReference()->willReturn(false);
+        $method->getCode()->willReturn('');
+
+        $code = $this->generate('CustomClass', $class);
+
+        $expected = <<<'PHP'
+namespace  {
+class CustomClass extends \stdClass implements  {
+
+public  function foo(): \Foo&\Bar|string {
```
stof, line 353: This is not a valid DNF type in the code: https://3v4l.org/aivA5 (I just removed the `implements` that is invalid code, as our ClassCodeGenerator apparently does not support the case of an empty list of interface, which is not an issue for Prophecy itself as we always add the ProphecySubjectInterface)

spec/Prophecy/Doubler/Generator/ClassCodeGeneratorSpec.php
```diff
@@ -310,6 +313,47 @@ public  function foo(): int|string|null {
...
 
+}
+}
+PHP;
+        $expected = strtr($expected, array("\r\n" => "\n", "\r" => "\n"));
+
+        $code->shouldBe($expected);
+    }
+
+    function it_generates_proper_code_for_intersection_return_types(
```
stof, line 324: I suggest having separate tests for an intersection return type (what the current name says) and for a DNF type (what it does)

Nek-: done!

src/Prophecy/Doubler/Generator/Node/Type/SimpleType.php
Nek-: Definitely true, I removed them!

src/Prophecy/Doubler/Generator/ClassCodeGenerator.php
```diff
@@ -78,16 +83,36 @@ private function generateMethod(Node\MethodNode $method): string
...
+
+    private function generateSubType(TypeInterface $type): string
+    {
+        if ($type instanceof SimpleType) {
+            return $type->getType();
         }
+
+        if ($type instanceof UnionType) {
+            return join('|', array_map(
+                fn (TypeInterface $type) => $this->generateSubType($type),
```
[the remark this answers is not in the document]
Nek-: Right, and fixed!

stof (commented):
The general approach looks good. I still recommend changing the way we manage the source code representation of a given type.
Please start working on the deprecation layer as well.

src/Prophecy/Doubler/Generator/Node/Type/BuiltinType.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Prophecy\Doubler\Generator\Node\Type;
+
+/**
+ * @final
```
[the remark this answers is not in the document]
Nek-: It's also for its own spec

stof:
@Nek- that's what I detailed more in https://github.com/phpspec/prophecy/pull/637#discussion_r2137978897 and https://github.com/phpspec/prophecy/pull/637#discussion_r2137972781

Nek-:
@stof I'm still working on this!

Removing deprecated usages is a new mini challenge and made me discover some features I forgot on the way, I will add according tests as well.

But I have an issue with triggering deprecation notice. It does makes the phpspec test suite fail, I can't make it pass with the trigger working. (I guess I'm not the first one hitting this issue since other deprecated methods do not throw deprecation error)
Is it ok to you if I push those deprecations commented?

Nek-:
> reading the code of phpspec, I think the pattern used in `symfony/deprecation-contracts` (triggering a _silenced_ deprecation) would work in phpspec (phpspec itself won't report it, but it will benefit people using phpunit with a config reporting them)

Awesome. I will use this. Thanks!

src/Prophecy/Doubler/Generator/ClassMirror.php
```diff
@@ -158,14 +164,22 @@ private function reflectMethodToNode(ReflectionMethod $method, Node\ClassNode $c
             $node->setReturnsReference();
         }
 
+        $returnReflectionType = null;
         if ($method->hasReturnType()) {
-            \assert($method->getReturnType() !== null);
-            $returnTypes = $this->getTypeHints($method->getReturnType(), $method->getDeclaringClass(), $method->getReturnType()->allowsNull());
-            $node->setReturnTypeNode(new ReturnTypeNode(...$returnTypes));
+            $returnReflectionType = $method->getReturnType();
+            \assert($returnReflectionType !== null);
```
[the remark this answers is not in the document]
Nek-: Haha indeed nice catch. 🎯

src/Prophecy/Doubler/Generator/Node/Type/UnionType.php
```diff
@@ -0,0 +1,119 @@
...
+
+        // To detect duplicates
+        $typeStrings = [];
+
+        foreach ($this->types as $type) {
+            if ($type instanceof UnionType) {
+                throw new DoubleException('Union types cannot contain other unions.');
+            }
+            if ($type instanceof IntersectionType) {
+                $typeStrings[] = implode('inter-', array_map(fn(SimpleType $type) => $type->getType(), $type->getTypes()));
```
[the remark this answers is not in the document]
stof: I would use `&`. It makes more sense.

src/Prophecy/Doubler/Generator/Node/TypeNodeAbstract.php
stof: yes, `new ReturnTypeNode(null)` should work. Otherwise, we won't be able to remove the BC layer in the next version anyway (where we don't want to have a variadic argument anymore)

src/Prophecy/Doubler/Generator/Node/Type/BuiltinType.php
```diff
@@ -2,13 +2,19 @@
 
 namespace Prophecy\Doubler\Generator\Node\Type;
 
-/**
- * @final
```
stof: Why removing `@final` there (without making it actually final) ?

composer.json
```diff
@@ -22,7 +22,8 @@
         "phpdocumentor/reflection-docblock": "^5.2",
         "sebastian/comparator":              "^3.0 || ^4.0 || ^5.0 || ^6.0 || ^7.0",
         "doctrine/instantiator":             "^1.2 || ^2.0",
-        "sebastian/recursion-context":       "^3.0 || ^4.0 || ^5.0 || ^6.0 || ^7.0"
+        "sebastian/recursion-context":       "^3.0 || ^4.0 || ^5.0 || ^6.0 || ^7.0",
+        "symfony/deprecation-contracts":     "^3.1"
```
stof, line 26: I suggest using `^2.5 || ^3.1` (or maybe `3.6`). Version 2 of `symfony/deprecation-contracts` 15% of installations according to the packagist.org stats, and there is no reason to prevent those projects from using an uptodate version of prophecy (of course, they should probably upgrade, but that's a separate topic and prophecy alone won't be enough to push that, so better not make them use old version).

spec/Prophecy/Argument/Token/ArrayEntryTokenSpec.php
```diff
@@ -83,8 +83,8 @@ function it_throws_exception_during_scoring_of_array_accessible_object_if_key_is
     ) {
         $key->__toString()->willReturn('any_token');
         $this->beConstructedWith($key, $value);
-        $errorMessage = 'You can only use exact value tokens to match key of ArrayAccess object'.PHP_EOL.
-                        'But you used `any_token`.';
+        $errorMessage = 'You can only use exact value tokens to match key of ArrayAccess object'.PHP_EOL
+                        .'But you used `any_token`.';
```
stof, line 87: I suggest rebasing on top of latest master, as those changes to existing code have been applied.

src/Prophecy/Doubler/Generator/Node/Type/ObjectType.php
```diff
@@ -0,0 +1,23 @@
...
+class ObjectType extends SimpleType
+{
+    public function __construct(string $type)
+    {
+        parent::__construct($this->prefixWithNsSeparator($type));
+    }
+
+    private function prefixWithNsSeparator(string $type): string
+    {
+        // Avoid double-prefixing if already prefixed
```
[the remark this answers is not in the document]
stof: I'm unmarking this as resolved, because you kept this logic (without any answer about why my comment got ignored)

src/Prophecy/Doubler/Generator/Node/Type/SimpleType.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Prophecy\Doubler\Generator\Node\Type;
+
+abstract class SimpleType implements TypeInterface
+{
+    public function __construct(private readonly string $type) {}
```
stof, line 7: I would turn `SimpleType` into an interface, moving this constructor the constructor to ObjectType and BuiltinType, which would allow ObjectType to describe its constructor as expecting `class-string`

src/Prophecy/Doubler/Generator/Node/TypeNodeAbstract.php
```diff
@@ -2,63 +2,202 @@
...
-    /** @var array<string, string> */
-    protected $types = [];
+    // null means no type, NOT BuiltInType("null")
+    private ?TypeInterface $type = null;
 
-    public function __construct(string ...$types)
+    /**
+     * @param string|TypeInterface ...$types
+     */
+    public function __construct(string|TypeInterface ...$types)
```
stof, line 21: The constructor signature should be public function __construct(string|TypeInterface|null $type = null, string ...$types)`, so that only the first argument (the only one in the new API) supports `TypeInterface|null`. The legacy variadic parameters support only strings.
This will even simplify the BC layer, as you won't need to check the type of the extra arguments.

src/Prophecy/Doubler/Generator/Node/TypeNodeAbstract.php
```diff
@@ -2,63 +2,201 @@
...
+            $this->type = null;
+        } elseif (count($types) === 1 && is_string($types[0])) {
+            // BC Layer for usage with strings
+            trigger_deprecation(
+                'phpspec/prophecy',
+                '1.23',
+                'Instanciating node type with a string type will not be supported in the future, use a TypeInterface instance instead.',
+            );
+            if ($this->isBuiltIn($types[0])) {
+                $this->type = new BuiltinType($types[0]);
```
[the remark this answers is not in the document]
stof: this does not seem to be done here (you did it on like 55 for the case passing multiple arguments, but not here)

stof (changes requested):

src/Prophecy/Doubler/Generator/Node/Type/ObjectType.php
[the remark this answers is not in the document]
stof: If the ObjectType is implemented as expecting `class-string`, we won't care about a type starting with `\`, as that won't happen. So `__toString` should just be `return '\\'.$this->type;`

And that's also solving the case of `equals` which currently treat `new ObjectType('stdClass')` and `new ObjectType('\stdClass')` as not equal.

src/Prophecy/Doubler/Generator/ClassMirror.php
```diff
@@ -233,52 +246,86 @@ private function getDefaultValue(ReflectionParameter $parameter)
...
+                if ($innerReflectionType instanceof ReflectionIntersectionType) {
+                    /** @var IntersectionType $intersection */
+                    $intersection = $this->createTypeFromReflection($innerReflectionType, $declaringClass);
+                    $innerTypes[] = $intersection;
+                    continue;
+                }
+                $name = $this->resolveTypeName($innerReflectionType->getName(), $declaringClass);
+                if ($innerReflectionType->isBuiltin() || $name === 'static') {
+                    $innerTypes[] = new BuiltinType($name);
+                } elseif ($name === 'self') {
```
stof, line 278: This is dead code. `resolveTypeName` will never return `self` as it resolves it.

src/Prophecy/Doubler/Generator/Node/Type/UnionType.php
```diff
@@ -0,0 +1,116 @@
...
+
+        // To detect duplicates
+        $typeStrings = [];
+
+        foreach ($this->types as $type) {
+            if ($type instanceof UnionType) {
+                throw new DoubleException('Union types cannot contain other unions.');
+            }
+            if ($type instanceof IntersectionType) {
+                $typeStrings[] = implode('&', array_map(fn(SimpleType $type) => $type->getType(), $type->getTypes()));
```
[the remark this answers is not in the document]
stof: I mean using `$typeStrings[] = (string) $type;`, which produces a non-ambiguous type representation while being very simple code relying on existing logic.

Btw, `$typeName` below could be changed that way as well.

src/Prophecy/Doubler/Generator/Node/TypeNodeAbstract.php
```diff
@@ -2,63 +2,198 @@
...
+                if (!is_string($type)) {
+                    throw new DoubleException('Building a TypeNode with string is deprecated. Mixing strings and type object is not allowed.');
+                }
+
+                if ($this->isBuiltIn($type)) {
+                    $type = new BuiltinType($this->normalizeBuiltinType($type));
+                } else {
+                    /** @var class-string $typeName */
+                    $typeName = $type;
+                    $type = new ObjectType($this->removePrefixNsSeparator($typeName));
```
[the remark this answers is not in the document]
Nek-: oops, this is fixed 😅

stof (approved):

Nek-:
@stof done!

stof:
It took time, but this is finally merged. Thanks a lot @Nek- !

julienbornstein:
Thanks @Nek- (nice to see you again)

Shivoham:
Thanks @Nek- !

PR #639, opened 2025-04-29
Bump to PHP 7.4+ and upgrade to PHPStan v2

Jean85:
@MarcelloDuarte this is the PR that I promised. I'll self-review to add more context to each change.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,31 +1,79 @@
 parameters:
 	ignoreErrors:
 		-
-			message: "#^Parameter \\#1 \\$value of function strval expects bool\\|float\\|int\\|resource\\|string\\|null, mixed given\\.$#"
+			message: '#^Parameter \#1 \$value of function strval expects bool\|float\|int\|resource\|string\|null, mixed given\.$#'
+			identifier: argument.type
 			count: 2
 			path: src/Prophecy/Argument/Token/ExactValueToken.php
 
 		-
-			message: "#^Method Prophecy\\\\Doubler\\\\CachedDoubler\\:\\:createDoubleClass\\(\\) should return class\\-string\\<Prophecy\\\\Doubler\\\\DoubleInterface&T of object\\> but returns class\\-string\\.$#"
+			message: '#^Method Prophecy\\Comparator\\ProphecyComparator\:\:assertEquals\(\) has parameter \$processed with no value type specified in iterable type array\.$#'
+			identifier: missingType.iterableValue
+			count: 1
+			path: src/Prophecy/Comparator/ProphecyComparator.php
+
+		-
+			message: '#^Method Prophecy\\Doubler\\CachedDoubler\:\:createDoubleClass\(\) should return class\-string\<Prophecy\\Doubler\\DoubleInterface&T of object\> but returns class\-string\.$#'
+			identifier: return.type
 			count: 1
 			path: src/Prophecy/Doubler/CachedDoubler.php
 
 		-
-			message: "#^Method Prophecy\\\\Doubler\\\\Doubler\\:\\:createDoubleClass\\(\\) should return class\\-string\\<Prophecy\\\\Doubler\\\\DoubleInterface&T of object\\> but returns class\\-string\\.$#"
+			message: '#^Method Prophecy\\Doubler\\Doubler\:\:createDoubleClass\(\) should return class\-string\<Prophecy\\Doubler\\DoubleInterface&T of object\> but returns class\-string\.$#'
+			identifier: return.type
 			count: 1
 			path: src/Prophecy/Doubler/Doubler.php
 
 		-
-			message: "#^Parameter \\#1 \\$callback of function array_map expects \\(callable\\(ReflectionIntersectionType\\|ReflectionNamedType\\|string\\)\\: mixed\\)\\|null, Closure\\(string\\)\\: string given\\.$#"
+			message: '#^Call to function method_exists\(\) with ''ReflectionClass'' and ''isReadOnly'' will always evaluate to true\.$#'
+			identifier: function.alreadyNarrowedType
 			count: 1
 			path: src/Prophecy/Doubler/Generator/ClassMirror.php
 
 		-
-			message: "#^Parameter \\#1 \\$callback of function array_map expects \\(callable\\(ReflectionIntersectionType\\|ReflectionNamedType\\)\\: mixed\\)\\|null, Closure\\(ReflectionNamedType\\)\\: string given\\.$#"
+			message: '#^Call to function method_exists\(\) with ReflectionMethod and ''hasTentativeReturnT…'' will always evaluate to true\.$#'
+			identifier: function.alreadyNarrowedType
+			count: 1
+			path: src/Prophecy/Doubler/Generator/ClassMirror.php
+
+		-
+			message: '#^Parameter \#1 \$callback of function array_map expects \(callable\(ReflectionIntersectionType\|ReflectionNamedType\|string\)\: mixed\)\|null, Closure\(string\)\: string given\.$#'
+			identifier: argument.type
+			count: 1
+			path: src/Prophecy/Doubler/Generator/ClassMirror.php
+
+		-
+			message: '#^PHPDoc tag @var with type static\(Prophecy\\Doubler\\LazyDouble\<U of object\>\) is not subtype of native type \$this\(Prophecy\\Doubler\\LazyDouble\<T of object\>\)\.$#'
+			identifier: varTag.nativeType
+			count: 1
+			path: src/Prophecy/Doubler/LazyDouble.php
+
+		-
+			message: '#^Property Prophecy\\Doubler\\LazyDouble\:\:\$class \(ReflectionClass\<T of object\>\|null\) does not accept ReflectionClass\<U of object\>\.$#'
+			identifier: assign.propertyType
+			count: 1
+			path: src/Prophecy/Doubler/LazyDouble.php
+
+		-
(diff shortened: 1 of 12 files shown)


phpstan-baseline.neon
```diff
@@ -1,31 +1,79 @@
 parameters:
 	ignoreErrors:
 		-
-			message: "#^Parameter \\#1 \\$value of function strval expects bool\\|float\\|int\\|resource\\|string\\|null, mixed given\\.$#"
+			message: '#^Parameter \#1 \$value of function strval expects bool\|float\|int\|resource\|string\|null, mixed given\.$#'
+			identifier: argument.type
```
Jean85, line 5: Multiple lines here are changed just because PHPStan v2 does no longer double-escape, and adds the identifier of the error. I'll comment only on new ignored issues.

phpstan-baseline.neon
```diff
@@ -1,31 +1,79 @@
...
 		-
-			message: "#^Method Prophecy\\\\Doubler\\\\CachedDoubler\\:\\:createDoubleClass\\(\\) should return class\\-string\\<Prophecy\\\\Doubler\\\\DoubleInterface&T of object\\> but returns class\\-string\\.$#"
+			message: '#^Method Prophecy\\Comparator\\ProphecyComparator\:\:assertEquals\(\) has parameter \$processed with no value type specified in iterable type array\.$#'
+			identifier: missingType.iterableValue
+			count: 1
+			path: src/Prophecy/Comparator/ProphecyComparator.php
+
+		-
+			message: '#^Method Prophecy\\Doubler\\CachedDoubler\:\:createDoubleClass\(\) should return class\-string\<Prophecy\\Doubler\\DoubleInterface&T of object\> but returns class\-string\.$#'
+			identifier: return.type
```
Jean85, line 17: This is a method extending its parent which is generic. It seems that the nested-double generic isn't working, we could report this upstream to PHPStan

phpstan-baseline.neon
```diff
@@ -1,31 +1,79 @@
...
+			message: '#^Call to function method_exists\(\) with ''ReflectionClass'' and ''isReadOnly'' will always evaluate to true\.$#'
+			identifier: function.alreadyNarrowedType
+			count: 1
+			path: src/Prophecy/Doubler/Generator/ClassMirror.php
+
+		-
+			message: '#^Call to function method_exists\(\) with ReflectionMethod and ''hasTentativeReturnT…'' will always evaluate to true\.$#'
+			identifier: function.alreadyNarrowedType
+			count: 1
+			path: src/Prophecy/Doubler/Generator/ClassMirror.php
```
Jean85, line 37: These two are due to the fact that those reflection methods are present only from PHP 8.1. These workaround go away as soon as we bump the minimum PHP version.

phpstan-baseline.neon
```diff
@@ -1,31 +1,79 @@
...
+			message: '#^Property Prophecy\\Doubler\\LazyDouble\:\:\$class \(ReflectionClass\<T of object\>\|null\) does not accept ReflectionClass\<U of object\>\.$#'
+			identifier: assign.propertyType
+			count: 1
+			path: src/Prophecy/Doubler/LazyDouble.php
+
+		-
+			message: '#^Property Prophecy\\Doubler\\LazyDouble\:\:\$class \(ReflectionClass\<T of object\>\|null\) is never assigned ReflectionClass\<T of object\> so it can be removed from the property type\.$#'
+			identifier: property.unusedType
+			count: 1
+			path: src/Prophecy/Doubler/LazyDouble.php
```
Jean85, line 61: These 3 are all from the same issue, the setParentClass method is very hard to map against PHPStan because it changes the type of the generic of the instance itself.

phpstan-baseline.neon
```diff
@@ -1,31 +1,79 @@
...
+
+		-
+			message: '#^Call to function method_exists\(\) with ReflectionMethod and ''hasTentativeReturnT…'' will always evaluate to true\.$#'
+			identifier: function.alreadyNarrowedType
+			count: 1
+			path: src/Prophecy/Prophecy/MethodProphecy.php
```
Jean85, line 67: Again, 8.1+ reflection feature...

src/Prophecy/Argument/Token/ArrayEntryToken.php
```diff
@@ -60,9 +60,8 @@ public function scoreArgument($argument)
 
         $keyScores = array_map(array($this->key,'scoreArgument'), array_keys($argument));
         $valueScores = array_map(array($this->value,'scoreArgument'), $argument);
-        /** @var callable(int|false, int|false): (int|false) $scoreEntry */
-        $scoreEntry = function ($value, $key) {
-            return $value && $key ? min(8, ($key + $value) / 2) : false;
+        $scoreEntry = static function ($value, $key) {
+            return $value && $key ? (int) min(8, ($key + $value) / 2) : false;
```
Jean85, line 64: Actually casting to int instead of var-typing

src/Prophecy/Doubler/Generator/ClassMirror.php
```diff
@@ -281,6 +281,6 @@ function (string $type) use ($class) {
             $types[] = 'null';
         }
 
-        return $types;
+        return array_values($types);
```
Jean85, line 284: Expecting to return a list

src/Prophecy/Exception/Prediction/UnexpectedCallsCountException.php
```diff
@@ -22,7 +22,7 @@ class UnexpectedCallsCountException extends UnexpectedCallsException
      * @param string         $message
      * @param MethodProphecy $methodProphecy
      * @param int            $count
-     * @param list<Call>     $calls
+     * @param array<Call>     $calls
```
Jean85, line 25: Inputing any array, casting to list below

src/Prophecy/Exception/Prediction/UnexpectedCallsException.php
```diff
@@ -17,18 +17,19 @@
...
      * @param MethodProphecy $methodProphecy
-     * @param list<Call>     $calls
+     * @param array<Call>     $calls
      */
     public function __construct($message, MethodProphecy $methodProphecy, array $calls)
     {
         parent::__construct($message, $methodProphecy);
 
-        $this->calls = $calls;
+        $this->calls = array_values($calls);
```
Jean85, line 32: Same here, casting to list (previous was extending class)

src/Prophecy/Prediction/CallbackPrediction.php
```diff
@@ -48,7 +48,7 @@ public function check(array $calls, ObjectProphecy $object, MethodProphecy $meth
     {
         $callback = $this->callback;
 
-        if ($callback instanceof Closure && method_exists('Closure', 'bind') && (new ReflectionFunction($callback))->getClosureThis() !== null) {
+        if ($callback instanceof Closure && (new ReflectionFunction($callback))->getClosureThis() !== null) {
```
Jean85, line 51: `bind` is present since PHP 5.4, we can ditch this check.

src/Prophecy/Promise/CallbackPromise.php
```diff
@@ -49,7 +49,7 @@ public function execute(array $args, ObjectProphecy $object, MethodProphecy $met
     {
         $callback = $this->callback;
 
-        if ($callback instanceof Closure && method_exists('Closure', 'bind') && (new ReflectionFunction($callback))->getClosureThis() !== null) {
+        if ($callback instanceof Closure && (new ReflectionFunction($callback))->getClosureThis() !== null) {
```
Jean85, line 52: `bind` is present since PHP 5.4, we can ditch this check.

src/Prophecy/Util/ExportUtil.php
```diff
@@ -193,6 +190,7 @@ protected static function recursiveExport(&$value, $indentation, $processed = nu
 
                 $values = "\n".$values.$whitespace;
             }
+            \assert(\is_object($value));
```
Jean85, line 193: This is just a change in how PHPStan detect types; better but not completely.

Jean85:
@MarcelloDuarte I've bumped to PHP 7.4 as discussed. 

I've also fixed the PHPStan analysis, but I saw that the output depends on the PHP version it runs on. It seems acceptable for now.

I also added a changelog entry, so that a 1.22 should be released with this bump. Let me know if this is ok or if it needs more work.

src/Prophecy/Util/ExportUtil.php
```diff
@@ -147,7 +147,6 @@ protected static function recursiveExport(&$value, $indentation, $processed = nu
                 return 'Array &'.$key;
             }
 
-            \assert(\is_array($value));
```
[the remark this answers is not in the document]
MarcelloDuarte: It's some sort of circular dependency around stan. That wasn't my question though... Is this mergeable? Do we need to keep support PHP versions in EOL state to keep the next tag within minor release?

src/Prophecy/Exception/Prediction/UnexpectedCallsCountException.php
stof: do we need this, or should we use a more precise type in our CallCenter (which is probably already using lists) ?

src/Prophecy/Util/ExportUtil.php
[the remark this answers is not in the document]
stof: > Do we need to keep supporting PHP versions in EOL state to keep the next tag within minor release?

As long as we update our `php` requirement properly in the composer.json, dropping support for some PHP versions is not a BC break (as composer will still respect that requirement when resolving semver ranges). That's still something that deserves a minor version IMO (it is not a bugfix), but it does not require a major version bump as far as Composer is concerned (Symfony itself has a policy to be stricter for the fullstack framework, but that's for other reasons)

PR #644, opened 2025-05-07
Upgrade PHP minimum version to 8.2

### LINKED ISSUE #431: Drop PHP 5.x/7.0 support
Both PHPUnit and PhpSpec regularly drop support for legacy PHP versions. Is it time for prophecy to do the same?

I suggest dropping any unmaintained PHP version and adopt a practice of doing it regularly.
Benefits of supporting less platforms would include less maintenance or being able to use more recent language features.

I'd be happy to help.

Nek-:
I started to work on #637 but supporting old versions of PHP is a lot of work and adds a layer we do not really need anymore since those versions are not supported anymore by PHP itself.

This is why I'm proposing here a PR that bump the minimum version of PHP.

https://www.php.net/supported-versions

Side note: _8.1 is also removed because a lot changed between 8.1 and 8.2 about types and the end of its support also comes soon enough._

Closes #431

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,11 +15,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: ["7.4", "8.0", "8.1", "8.2", "8.3"]
+        php: ["8.2", "8.3"]
         composer-flags: [ "" ]
         experimental: [ false ]
         include:
-          - php: 7.4
+          - php: 8.2
             composer-flags: "--prefer-lowest"
           - php: "8.4" # TODO move that to a normal job once phpspec supports PHP 8.4
             composer-flags: "--ignore-platform-req=php+"
@@ -59,7 +59,7 @@ jobs:
       - name: Set up PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: "8.2"
+          php-version: "8.3"
           coverage: none
       - name: Install dependencies
         run: COMPOSER_ROOT_VERSION=dev-master composer update --no-scripts
@@ -73,7 +73,7 @@ jobs:
       - name: Set up PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: "8.2"
+          php-version: "8.3"
           coverage: none
       - name: Install dependencies
         run: COMPOSER_ROOT_VERSION=dev-master composer update --no-scripts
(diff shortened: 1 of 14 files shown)


.github/workflows/build.yml
```diff
@@ -15,11 +15,11 @@ jobs:
...
       matrix:
-        php: ["7.4", "8.0", "8.1", "8.2", "8.3"]
+        php: ["8.2", "8.3", "8.4"]
         composer-flags: [ "" ]
         experimental: [ false ]
         include:
-          - php: 7.4
+          - php: 8.2
             composer-flags: "--prefer-lowest"
           - php: "8.4" # TODO move that to a normal job once phpspec supports PHP 8.4
```
stof, line 24: This job should be removed as you solved the TODO.

stof (commented):
I suggest reducing the range of PHPUnit versions in the dev requirements, as old versions are not needed anymore for our CI.

Nek-:
I updated the branch. Can you have a look to the phpstan error ? I don't understand how to fix it.

Nek-:
An approval for the Ci would be super-nice @stof :)

.github/workflows/build.yml
stof: This job is not about a min version at all. It is related to the fact that it was impossible to test 8.4 as a normal job not having the special composer flag to ignore the upper bound of the PHP requirement (because phpspec was not yet declared compatible with PHP 8.4).
The fact that you were able to add 8.4 as a normal job a few lines above is exactly the proof that this todo is resolved (well, partially: you copied it as a normal job instead of moving it, which is why I ask for this removal)

stof: OK, I see that you reverted that part.

.github/workflows/build.yml
```diff
@@ -59,7 +59,7 @@ jobs:
       - name: Set up PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: "8.2"
+          php-version: "8.4"
```
stof, line 62: use PHP 8.3 here, as dependencies cannot be installed on 8.4 without special flags because of phpspec.

stof:
which error ?

Nek-:
> which error ?

Errors that seems to be only by running phpstan in PHP 8.4:

image

stof:
I'll deal with that when changing our phpstan job to 8.4 in the future (once we can do that in a simple way)

Nek-:
All good for me, this can be merged then :) .

PR #646, opened 2025-07-25
Return `$this` for non-static methods with `self` or `static` return types for dummies

jdreesen:
Fixes the following error for dummies with fluent interface:

> TypeError : Double\SomeClass\P12::someMethod(): Return value must be of type Double\SomeClass\P12, null returned


stof:
This should be implemented in the MethodProphecy when guessing the default return value (like for other types) instead of changing the generated code.

jdreesen:
I'll take a look at that. Thanks for the feedback.

stof:
apparently, you already has a look at it in https://github.com/phpspec/prophecy/pull/645

jdreesen:
But as far as I remember, the change in #645 was not enough to make it work :thinking:

jdreesen:
I think the problem with #645 not being sufficient was that `MethodProphecy` is not used at all unless the method is explicitly specified. So you would have to explicitly list all fluent methods for it to work.

PR #653, opened 2025-10-09
Add missing docblocks

Nek-:
I figured out that I didn't add the docblocks in my previous PR #637 . This is a fix for it. (feel free to close it if it does not matter!)


stof:
Given that existing files have it, I think it makes sense to have them.

However, it would also make sense to update the php-cs-fixer config to make it aware of this license header so that it actually enforces it.

PR #659, opened 2025-11-14
Support readonly

### LINKED ISSUE #586: Mocking PHP 8.2 readonly classes generated readonly double with untyped property and fails

Nek-:
This PR is supposed to fix #586 and replace #623 .

### DIFF
--- a/src/Prophecy/Doubler/Generator/Node/PropertyNode.php
+++ b/src/Prophecy/Doubler/Generator/Node/PropertyNode.php
@@ -1,0 +1,97 @@
+<?php
+
+/*
+ * This file is part of the Prophecy.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *     Marcello Duarte <marcello.duarte@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Prophecy\Doubler\Generator\Node;
+
+use Prophecy\Exception\InvalidArgumentException;
+
+/**
+ * Property node.
+ */
+class PropertyNode
+{
+    private $name;
+
+    /**
+     * @var string
+     *
+     * @phpstan-var 'public'|'private'|'protected'
+     */
+    private $visibility = 'public';
+
+    /**
+     * @var PropertyTypeNode
+     */
+    private $typeNode;
+
+    /**
+     * @param string $name
+     */
+    public function __construct(string $name)
+    {
+        $this->name = $name;
+        $this->typeNode = new PropertyTypeNode();
+    }
+
+    /**
+     * @return string
+     */
+    public function getName(): string
+    {
+        return $this->name;
+    }
+
+    /**
+     * @return PropertyTypeNode
+     */
+    public function getTypeNode(): PropertyTypeNode
+    {
+        return $this->typeNode;
+    }
+
+    /**
+     * @return void
+     */
+    public function setTypeNode(PropertyTypeNode $typeNode)
+    {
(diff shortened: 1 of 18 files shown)


Nek-:
@stof this PR is ready for review!

src/Prophecy/Doubler/Generator/Node/PropertyNode.php
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+namespace Prophecy\Doubler\Generator\Node;
+
+use Prophecy\Exception\InvalidArgumentException;
+
+/**
+ * Property node.
+ */
+class PropertyNode
```
stof: I suggest making this new class final (or at least `@final` if our specs require mocking it)

src/Prophecy/Doubler/Generator/Node/PropertyNode.php
```diff
@@ -0,0 +1,89 @@
...
+     */
+    public function getTypeNode(): PropertyTypeNode
+    {
+        return $this->typeNode;
+    }
+
+    /**
+     * @return void
+     */
+    public function setTypeNode(PropertyTypeNode $typeNode): void
```
stof: Do we actually need the PropertyNode to be mutable ? If no, I would remove the setters.

src/Prophecy/Doubler/Generator/Node/PropertyNode.php
```diff
@@ -0,0 +1,89 @@
...
+
+    /** @phpstan-var 'public'|'private'|'protected' */
+    private string $visibility;
+
+    private PropertyTypeNode $typeNode;
+
+    /**
+     * @phpstan-param 'public'|'private'|'protected' $visibility
+     */
+    public function __construct(string $name, string $visibility = 'public', PropertyTypeNode $typeNode = new PropertyTypeNode())
```
stof: instead of this `PropertyTypeNode` extending `TypeNodeAbstract`, I suggest working directly with a `TypeInterface|null` here (i.e. our modern API for types). We don't need to preserve BC with the `TypeNodeAbstract` system here, as this is new code.

src/Prophecy/Doubler/ObjectProphecyClosureContainer.php
```diff
@@ -0,0 +1,27 @@
...
+namespace Prophecy\Doubler;
+
+/**
+ * Container for the closure that can be used and modified in a read-only class.
+ *
+ * @internal
+ *
+ * @noinspection PhpUnused
+ */
+class ObjectProphecyClosureContainer
```
stof: I would make it final

Nek-:
feedback done @stof

Nek-:
ping @stof

PR #660, opened 2026-01-18
Add support for `phpdocumentor/reflection-docblock` v6

### LINKED ISSUE #661: make phpdocumentor/reflection-docblock:6.0 compatible
can you make it compatible to phpdocumentor/reflection-docblock:6.0.0
mainly for composer issues with upgrading to 6.0.0

mspirkov:
The package https://github.com/phpDocumentor/Reflection needs to be upgraded to `phpdocumentor/reflection-docblock` v6, but this is currently not possible because `phpspec/prophecy` requires version `^5.2`.

image

Closes #661

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
 
     "require": {
         "php":                               "8.2.* || 8.3.* || 8.4.* || 8.5.*",
-        "phpdocumentor/reflection-docblock": "^5.2",
+        "phpdocumentor/reflection-docblock": "^5.2 || ^6.0",
         "sebastian/comparator":              "^3.0 || ^4.0 || ^5.0 || ^6.0 || ^7.0 || ^8.0",
         "doctrine/instantiator":             "^1.2 || ^2.0",
         "sebastian/recursion-context":       "^3.0 || ^4.0 || ^5.0 || ^6.0 || ^7.0 || ^8.0",
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -65,3 +65,9 @@ parameters:
 			identifier: offsetAccess.invalidOffset
 			count: 1
 			path: src/Prophecy/Util/ExportUtil.php
+
+		-
+			message: '#^Call to an undefined method phpDocumentor\\Reflection\\DocBlock\\Tags\\Method\:\:getArguments\(\)\.$#'
+			identifier: method.notFound
+			count: 1
+			path: src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php


stof:
Adding support for it seems to require more work than just changing the composer requirement.

mspirkov:
Yes. I see errors. I'll fix them.

mspirkov:
We also need this fix: https://github.com/phpspec/phpspec/pull/1503

src/Prophecy/Util/ExportUtil.php
```diff
@@ -78,6 +78,7 @@ public static function toArray($value)
         // above (fast) mechanism nor with reflection in Zend.
         // Format the output similarly to print_r() in this case
         if ($value instanceof \SplObjectStorage) {
+            /** @var \SplObjectStorage<object, mixed> $value */
```
mspirkov: After the release of PHPStan 2.1.34, a new error appeared. I suggest fixing it as follows.

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
```diff
@@ -83,7 +83,7 @@ public function apply(ClassNode $node)
                                     $argumentNode = new ArgumentNode($argument->getName());
                                     $methodNode->addArgument($argumentNode);
                                 }
-                            } else {
+                            } elseif (method_exists($tag, 'getArguments')) {
```
stof: Why do we need to add this if Reflection Docblock 5.4.0+ is already handled above ?

If Reflection Docblock 6.0+ has a third way of accessing those, it must be handled properly (and I would handle it first then, trying new APIs first)

src/Prophecy/Util/ExportUtil.php
stof: this should be reported to phpstan. In SplObjectStorage, keys are always objects, so it should not infer a wider type.

src/Prophecy/PhpDocumentor/ClassTagRetriever.php
```diff
@@ -50,6 +51,11 @@ public function getTagList(\ReflectionClass $reflectionClass)
             return $methods;
         } catch (\InvalidArgumentException $e) {
             return array();
+        } catch (CannotCreateTag $e) {
+            // In version 6.0, all deprecated methods have been removed from "ReflectionDocblock".
+            // Now, exceptions are thrown when trying to create a tag with a type and in other
+            // situations that previously did not throw exceptions.
```
stof: this looks suspicious to me. AFAICT, we should not be trying to create tags directly (and if we currently do, we should instead fix the root cause). In which case does this exception happen in a way that should be caught to ignore all tags ?

src/Prophecy/Util/ExportUtil.php
mspirkov: Yes, I agree, but what should I do here to fix this error? Ignore it?

stof: We could ignore it in the phpstan.neon config file, with a comment pointing to the issue in the phpstan issue tracker.

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
mspirkov: This appeared here due to static analysis errors. I made an explicit check to avoid ignoring this error.

image

src/Prophecy/PhpDocumentor/ClassTagRetriever.php
mspirkov: It's hard to say, because I don't use this package and I don't fully understand what's going on inside it. I'll take a closer look.

src/Prophecy/Doubler/ClassPatch/MagicCallPatch.php
stof: We put those in the phpstan ignore rules instead (as we know they are related to supporting multiple versions) instead of having a no-op `else` branch in case none of our checks are matching. This makes it easier to detect issues when trying to add support for a new version (tests would reach a non-working branch instead of reaching a no-op branch, and will report a failure).

It is known that writing code supporting multiple versions of a dependency often involve ignoring some static analysis errors.

src/Prophecy/PhpDocumentor/ClassTagRetriever.php
stof: what was the behavior without adding this catch statement ?

mspirkov: Tests fail due to exceptions: https://github.com/phpspec/prophecy/actions/runs/21114065815/job/60726462701

stof: Given that the static `create` method of the `Method` tag is not meant to be called anymore (the proper tag factory should be used instead), the actual root cause should be investigated (inspecting the stack trace to see where it is called). It might be an issue in the way we use the library (which then must be fixed) or an upstream bug. But silently ignoring all tags when a `CannotCreateTag` exception happen while this exception only happens when reaching an unsupported code path does not make sense (and would probably break support for magic methods)

mspirkov: I just don't understand how to see the stack trace when running `phpspec` tests

mspirkov: I've already figured it out :)

image

mspirkov: Yes, you're right. The problem is in the ReflectionDocblock. Thank you for looking closely!

mspirkov: I will fix the problem in ReflectionDocblock and come back here.

stof: next step would be debugging *which* tag is attempted to be created in the StandardTagFactory (and potentially also debugging what is the doc comment from which the phpdoc is parsed).

I suspect this is an upstream issue.

mspirkov: https://github.com/phpDocumentor/ReflectionDocBlock/pull/438

mspirkov: It seems that all we need to do is restart CI after `ReflectionDocBlock` release.

src/Prophecy/Util/ExportUtil.php
mspirkov: https://github.com/phpstan/phpstan/issues/13985

src/Prophecy/PhpDocumentor/ClassTagRetriever.php
mspirkov: https://github.com/phpDocumentor/ReflectionDocBlock/releases/tag/6.0.1

mspirkov:
@stof It looks like everything is ready here. All that's needed is a phpspec release with Symfony 8 support.

mspirkov:
I see a new release: https://github.com/phpspec/phpspec/releases/tag/8.2.0
I tested it locally, and it works.

image

mspirkov:
@stof It would be great if you could merge this. Then I could continue my work.

stof (approved):

PR #662, opened 2026-02-07
Allow PHPUnit 13

Jean85:
Draft because this is blocked by PHPSpec allowing the same.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,9 +20,9 @@
     "require": {
         "php":                               "8.2.* || 8.3.* || 8.4.* || 8.5.*",
         "phpdocumentor/reflection-docblock": "^5.2",
-        "sebastian/comparator":              "^3.0 || ^4.0 || ^5.0 || ^6.0 || ^7.0",
+        "sebastian/comparator":              "^3.0 || ^4.0 || ^5.0 || ^6.0 || ^7.0 || ^8.0",
         "doctrine/instantiator":             "^1.2 || ^2.0",
-        "sebastian/recursion-context":       "^3.0 || ^4.0 || ^5.0 || ^6.0 || ^7.0",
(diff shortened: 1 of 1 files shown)


Jean85:
#663 shows that this PR is green even when run with phpspec/phpspec#1510, which in turn allows PHPUnit 13: https://github.com/phpspec/prophecy/actions/runs/21784104191/job/62852657870#step:7:9

@stof can we merge this so we break the circular dependency between PRs with PHPSpec?

composer.json
```diff
@@ -20,17 +20,17 @@
...
+        "sebastian/recursion-context":       "^3.0 || ^4.0 || ^5.0 || ^6.0 || ^7.0 || ^8.0",
         "symfony/deprecation-contracts":     "^2.5 || ^3.1"
     },
 
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^3.88",
         "phpspec/phpspec": "^6.0 || ^7.0 || ^8.0",
-        "phpstan/phpstan": "^2.1.13",
-        "phpunit/phpunit": "^11.0 || ^12.0"
+        "phpstan/phpstan": "^2.1.13, <=2.1.33",
```
stof: why adding this highest bound here ?

Jean85: As written in the commit message, due to https://github.com/phpstan/phpstan/issues/13985 hitting us. The fix is merged but not released yet, once it is we can bump the constraint to that version.

stof: I submitted https://github.com/phpspec/prophecy/pull/665 with a better fix, preventing only the broken release but not future ones.

stof: ah, my PR is flawed. I forgot that 2.1.34 is not the latest version.

stof: fixed in https://github.com/phpspec/prophecy/pull/666

PR #667, opened 2026-02-21
Catch Throwable instead of Exception in CallCenter

### LINKED ISSUE #595: When a mocked method throws a `Throwable` the call to the method is not recorded
Given this code:
```php
$this->prophesize(Publisher::class)
	->processChanges($interval)
		->shouldBeCalled()
		->willThrow(new TypeError())
		->getObjectProphecy()
	->reveal();
```
when the `processChanges` method is called, it throws the specified `TypeError`, but the call is not recorded.

The reason for this are that:
1. `CallCenter` catches `Exception`, not `Throwable`
2. `Call` class requires an `?Exception` in the constructor, not `Throwable`.

IMO it is a perfectly valid assumption that a piece of code can cause a type error. It's not often, but it is a valid case specifically with generated code. I checked that if I just change `Exception` to `Throwable` in the two places, it works. Not sure if that breaks some other cases though.

ranpafin:
When a promise throws a Throwable that is not an Exception (e.g. TypeError),
CallCenter does not catch it, so the call is never recorded.

This changes the catch block in CallCenter and the type hint in Call from
Exception to Throwable, aligning them with ThrowPromise which already
accepts Throwable.

Closes #595


Nek-:
Prophecy is a highly opiniated framework. I think not catching throwable is part of it since you should never throw PHP errors.

Issue #669, opened 2026-04-06
ComparisonFailure throwable need to be updated


mayottecummings-seedbox:
In phpspec/prophecy/src/Prophecy/Comparator/ClosureComparator.php::assertEquals(), inside the IF, we have a throwable, but with the newest version of SebastianBergmann\Comparator, the constructor is broken. The constructor try to use the string `'all closures are different if not identical'` as an Interger.

Plus, the parameter "false" is not use anymore.

The possible solution could be to just remove the IF.

stof:
Fixed in #671

PR #670, opened 2026-04-08
Fix ComparisonFailure constructor compatibility with sebastian/comparator v8

jorgsowa:
`sebastian/comparator` < 5 has a 6-param constructor with $identical (bool) as 5th param, while >= 8 also has 6 params but with $contextLines (int) as 6th param. The previous check only looked at parameter count (>= 6), causing wrong types to be passed on v8. Now also checks the parameter name to distinguish the two signatures.

Fixes #[669](https://github.com/phpspec/prophecy/issues/669)


stof:
Closing this in favor of #671 in which I implemented a better fix, to actually support the new feature of `sebastian/comparator` 8.1+.

jorgsowa:
Thanks!

I think that workaround for sebastian/comparator < 5 may be even removed from the code. I didn't think about it previously.

stof:
@jorgsowa it cannot be removed as long as we don't remove support for using Prophecy alongside those versions of `sebastian/comparator` (which ultimately restrict the compatibility with phpunit): https://github.com/phpspec/prophecy/blob/09c2e5949d676286358a62af818f8407167a9dd6/composer.json#L23

PR #673, opened 2026-07-05
PHP 8.6 fixes

### LINKED ISSUE #672: PHP 8.6 - Returning a value from a destructor is deprecated
When a doubled class specifies a `__destruct` method, `ClassCodeGenerator` currently builds

```
<?php
namespace Double\xxxx {
class P1 extends \xxxx implements \Prophecy\Doubler\DoubleInterface, \Prophecy\Prophecy\ProphecySubjectInterface, \Prophecy\Doubler\Generator\ReflectionInterface {

...

public  function __destruct() {
}

...

}
```

and this will throw a deprecation in PHP 8.6.

It's due to [PHP RFC: Deprecate returning values from __construct() and __destruct()](https://wiki.php.net/rfc/deprecate-return-value-from-construct), that was committed in https://github.com/php/php-src/pull/21982.

andypost:
Fixes #672


src/Prophecy/Doubler/ClassPatch/ProphecySubjectPatch.php
```diff
@@ -48,11 +48,13 @@ public function apply(ClassNode $node)
...
+
+            if ('__construct' === $methodName) {
                 continue;
             }
 
-            if (!$method->getReturnTypeNode()->hasReturnStatement()) {
+            if ('__destruct' === $methodName || !$method->getReturnTypeNode()->hasReturnStatement()) {
                 $method->setCode(
                     '$this->getProphecy()->makeProphecyMethodCall(__FUNCTION__, func_get_args());'
                 );
```
neclimdul, line 60: This does make sense but I was wondering if the special case should be handled at the source. Digging down `hasReturnTypeStatement` I found its providing some convenience wrapping of [BetterReflection](https://github.com/ondrejmirtes/BetterReflection)'s logic. Their issues seem closed and I'm not sure where I'd start with a PR so this is probably a logical fix for now.

andypost: thanks for idea, need to dig it deeper

andypost: > special case should be handled at the source

exactly, PHP 8.6 constructors and destructors are symmetric: neither may return a value, so both must be proxied rather than return-typed so pushing it down will invert deps

